PR #1, opened 2023-02-06
Arm

sofiebudman:


.SysId/sysid.json
```diff
@@ -0,0 +1,10 @@
+{
+  "SysId": {
+    "Analysis Type": "General Mechanism",
+    "NetworkTables Settings": {
+      "serverTeam": "199"
+    },
+    "mode": "Client (NT4)",
+    "serverTeam": "199"
+  }
+}
```
CoolSpy3: This file should be deleted

src/main/java/org/carlmontrobotics/robotcode2023/Robot.java
```diff
@@ -28,6 +28,7 @@ public void robotPeriodic() {
     MotorErrors.printSparkMaxErrorMessages();
   }
 
+
```
CoolSpy3: Delete newline

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
```diff
@@ -13,12 +15,23 @@
...
+    }
+
     public static final class OI {
         public static final class Driver {
-            public static final int port = 0;
+            public static final int port = 1;
         }
         public static final class Manipulator {
-            public static final int port = 1;
+            public static final int port = 0;
```
CoolSpy3: Driver should be port `0`, Manipulator should be port `1`

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
```diff
@@ -13,12 +15,23 @@
...
+            public static final int cone = Button.kRightBumper.value;
+            public static final int cube = Button.kLeftBumper.value;
+            public static final int intake = Button.kA.value;
+            
+            public static final int cycleUp = 1;//FIXME use correct buttons
+            public static final int cycleDown = 2;
```
CoolSpy3: Use button constants

src/main/java/org/carlmontrobotics/robotcode2023/RobotContainer.java
```diff
@@ -4,25 +4,35 @@
...
   public RobotContainer() {
     configureButtonBindingsDriver();
     configureButtonBindingsManipulator();
   }
 
   private void configureButtonBindingsDriver() {}
-  private void configureButtonBindingsManipulator() {}
+  private void configureButtonBindingsManipulator() {
+    new JoystickButton(manipulatorController, Constants.OI.Manipulator.cycleUp).onTrue(new InstantCommand(()->arm.cycleUp()));
+    new JoystickButton(manipulatorController, Constants.OI.Manipulator.cycleDown).onTrue(new InstantCommand(()->arm.cycleDown()));
```
CoolSpy3: arm::cycleUp, arm::cycleDown

~~The commands should require Arm~~

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,171 @@
...
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
+import com.revrobotics.SparkMaxPIDController;
+import com.revrobotics.CANSparkMax;
+import com.revrobotics.RelativeEncoder;
+
+
+public class Arm extends SubsystemBase
+{
+  public CANSparkMax motor = MotorControllerFactory.createSparkMax(Constants.Arm.port, TemperatureLimit.NEO);
+  public RelativeEncoder motorLencoder = motor.getEncoder();
```
CoolSpy3: Make `private final`

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,171 @@
...
+  private double kV = 1.7912; //volts*secs/rad | extra velocity
+  private double kA = .15225; //volts*secs^2/rad | vacceleration
+  /// these are all units ^ , actual arm0.15225 speed is determined by values in .calculate
+  private double Kp = 2.2985;
+  private double Kd = 1.2943;
+  private double Ki = 0.0;
+  
+  private double setpoint = 2.2;
+  private double FFvelocity = 6;
+  private double FFaccel = 6;
```
CoolSpy3: Move to `Constants` and make `public static final`

If you need them to change via SmartDashboard, make the initial values constant and add the word default to the name.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,171 @@
...
+  private ArmFeedforward armFeed = new ArmFeedforward(kS, kG, kV, kA);
+  private PIDController pid = new PIDController(Kp, Ki, Kd);
+  
+  public double goalPos;
+
+  public enum ArmPreset {
+    INTAKE(0), MID(1), HIGH(2);
+    
+    public double value; //not static so SmartDashboard can touch [IMPORTANT TO KNOW!]
+    ArmPreset(double value) {
```
CoolSpy3: Make private

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,171 @@
...
+  private double FFaccel = 6;
+  private ArmFeedforward armFeed = new ArmFeedforward(kS, kG, kV, kA);
+  private PIDController pid = new PIDController(Kp, Ki, Kd);
+  
+  public double goalPos;
+
+  public enum ArmPreset {
+    INTAKE(0), MID(1), HIGH(2);
+    
+    public double value; //not static so SmartDashboard can touch [IMPORTANT TO KNOW!]
```
CoolSpy3: Add newline after

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,171 @@
...
+      return null;
+    }
+    public ArmPreset prev(){
+      switch (this) {
+        case INTAKE: return HIGH;
+        case MID: return INTAKE;
+        case HIGH: return MID;
+      }
+      return null;
+    }
```
CoolSpy3: This requires updating the methods if new values are added/removed. They also shouldn't be able to return `null`.

I recommend `return values()[(ordinal() + 1) % values().length]` and `return values()[(ordinal - 1 + values().length) % values().length]` (someone should also do a more thorough test that those expressions work.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,171 @@
...
+    
+  }
+
+  public Arm(){
+    motorLencoder.setPositionConversionFactor(2*Math.PI/100);
```
CoolSpy3: Add a gearing constant

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,171 @@
...
+    SmartDashboard.putNumber("ArmLencoderPos", motorLencoder.getPosition());
+    SmartDashboard.putString("ArmENUM", 
+    (closeSnappedArmPos() != null) ? closeSnappedArmPos().toString() : snappedArmPos().toString()
+    );
+    Kp = SmartDashboard.getNumber("KP", Kp);
+    Ki = SmartDashboard.getNumber("KI", Ki);
+    Kd = SmartDashboard.getNumber("KD", Kd);
+    pid.setP(Kp);
+    pid.setI(Ki);
+    pid.setD(Kd);
```
CoolSpy3: All of these (Except maybe ArmENUM) should probably be moved to LiveWindow

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  @Override
+  public void periodic() {
+    FFvelocity = SmartDashboard.getNumber("FF: Velocity", FFvelocity);
+    SmartDashboard.getNumber("KG", kG);
+    FFaccel = SmartDashboard.getNumber("FF: Acceleration", FFaccel);
+    goalPos = SmartDashboard.getNumber("GoalPosition", goalPos);
+    SmartDashboard.putNumber("ArmLencoderPos", motorLencoder.getPosition());
+    SmartDashboard.putString("ArmENUM",
```
CoolSpy3: camelCase Enum

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,171 @@
...
+import com.revrobotics.CANSparkMax;
+import com.revrobotics.RelativeEncoder;
+
+
+public class Arm extends SubsystemBase
+{
+  public CANSparkMax motor = MotorControllerFactory.createSparkMax(Constants.Arm.port, TemperatureLimit.NEO);
+  public RelativeEncoder motorLencoder = motor.getEncoder();
+
+  public double encoderErrorTolerance = .1;
```
CoolSpy3: Document that this is Radians

src/main/java/org/carlmontrobotics/robotcode2023/RobotContainer.java
```diff
@@ -4,25 +4,35 @@
...
+import org.carlmontrobotics.robotcode2023.subsystems.Arm;
+import org.carlmontrobotics.robotcode2023.subsystems.Arm.ArmPreset;
+
 import edu.wpi.first.wpilibj.Joystick;
 import edu.wpi.first.wpilibj.PowerDistribution;
 import edu.wpi.first.wpilibj.XboxController;
 import edu.wpi.first.wpilibj2.command.Command;
 import edu.wpi.first.wpilibj2.command.Commands;
+import edu.wpi.first.wpilibj2.command.InstantCommand;
+import edu.wpi.first.wpilibj2.command.button.JoystickButton;
```
CoolSpy3: Run Organize Imports

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,171 @@
...
+import org.carlmontrobotics.robotcode2023.Constants;
+
+import edu.wpi.first.wpilibj.smartdashboard.SmartDashboard;
+import edu.wpi.first.math.controller.ArmFeedforward;
+import edu.wpi.first.math.controller.PIDController;
+import edu.wpi.first.math.util.Units;
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
+import com.revrobotics.SparkMaxPIDController;
+import com.revrobotics.CANSparkMax;
+import com.revrobotics.RelativeEncoder;
```
CoolSpy3: Run organize imports

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,171 @@
...
+  private double kS = .17764; //volts | base speed
+  private double kG = 7.1181; //volts | gravity... something
+  private double kV = 1.7912; //volts*secs/rad | extra velocity
+  private double kA = .15225; //volts*secs^2/rad | vacceleration
+  /// these are all units ^ , actual arm0.15225 speed is determined by values in .calculate
+  private double Kp = 2.2985;
+  private double Kd = 1.2943;
+  private double Ki = 0.0;
+  
+  private double setpoint = 2.2;
```
CoolSpy3: Unused

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,171 @@
...
+    SmartDashboard.putNumber("FF: Acceleration", FFaccel);
+    SmartDashboard.putNumber("GoalPosition", goalPos);
+    SmartDashboard.putNumber("KG", kG);
+    SmartDashboard.putString("Debooog", "No.");
+    SmartDashboard.putString("ArmENUM", 
+    snappedArmPos().toString()
+    );
+    SmartDashboard.putNumber("KP", Kp);
+    SmartDashboard.putNumber("KI", Ki);
+    SmartDashboard.putNumber("KD", Kd);
```
CoolSpy3: See comment in `periodic()`

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,171 @@
...
+    pid.setTolerance(2.5,10);
```
CoolSpy3: Magic numbers, make these into constants

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,171 @@
...
+    double encoderPos = motorLencoder.getPosition();
+    
+    for(ArmPreset check : ArmPreset.values()){
+        if (Math.abs(check.value - encoderPos) < encoderErrorTolerance) {
+            return check;
+        }
+    }
+    //help something went wrong
+    return null;
+  }
```
CoolSpy3: Just store the target `ArmPreset` and return that

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,171 @@
...
+
+  public void cycleUp(){ 
+    // because most people won't remember/want to do this long function chain
+    SmartDashboard.putString("Debooog", "CycleUp");
+    SmartDashboard.putNumber("GoalPosition", 
+    closeSnappedArmPos() != null ? closeSnappedArmPos().next().value : snappedArmPos().next().value
```
CoolSpy3: See comments in `periodic()` and constants

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,171 @@
...
+            return check;
+        }
+    }
+    //help something went wrong
+    return null;
+  }
+
+  public void cycleUp(){ 
+    // because most people won't remember/want to do this long function chain
+    SmartDashboard.putString("Debooog", "CycleUp");
```
CoolSpy3: Use a more descriptive name

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,171 @@
...
+    for(ArmPreset check : ArmPreset.values()){
+        if (Math.abs(check.value - encoderPos) < encoderErrorTolerance) {
+            return check;
+        }
+    }
+    //help something went wrong
+    return null;
+  }
+
+  public void cycleUp(){
```
CoolSpy3: Make the method name more descriptive or add documentation

CoolSpy3: Forward call to `setPreset()`

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,171 @@
...
+  
+  public void cycleDown(){
+    SmartDashboard.putString("Debooog", "CycleDown");
+    SmartDashboard.putNumber("GoalPosition", 
+    closeSnappedArmPos() != null ? closeSnappedArmPos().prev().value : snappedArmPos().prev().value
+    );
+  }
+  
+  public void setPreset(ArmPreset preset){
+    SmartDashboard.putNumber("GoalPosition", preset.value);
```
CoolSpy3: See comments in `periodic()` and constants

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,171 @@
...
+  public void cycleUp(){ 
+    // because most people won't remember/want to do this long function chain
+    SmartDashboard.putString("Debooog", "CycleUp");
+    SmartDashboard.putNumber("GoalPosition", 
+    closeSnappedArmPos() != null ? closeSnappedArmPos().next().value : snappedArmPos().next().value
+    );
+    // if closeSnappedArmPos is working, swap based on it - otherwise use less accurate snapping
+  }
+  
+  public void cycleDown(){
```
CoolSpy3: See comments in cycleUp()

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
CoolSpy3: Simply using the next and prev values could request that the arm move a large rotation after a single driver input. this may not always be desirable. Not requesting any changes, but should be kept in mind.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,171 @@
...
+    if (Math.abs(difference) > encoderErrorTolerance){//even PID needs an acceptable error sometimes
+      //assuming calculate() is some sort of PID-esque thing
+      motor.setVoltage(armFeed.calculate(goalPos, FFvelocity, FFaccel)
+         + pid.calculate(motorLencoder.getPosition(), goalPos));
+      //motor.set(dir * armFeed.calculate(goalPos + Math.PI/2, FFvelocity, FFaccel));
+    } else {
+      //motor.set(dir * armFeed.calculate(goalPos + Math.PI/2, 0, 0));
+      motor.setVoltage(armFeed.calculate(goalPos, 0, 0)
+         + pid.calculate(motorLencoder.getPosition(), goalPos));
+    }
```
CoolSpy3: Make this into a default command

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
CoolSpy3: Change name to just `motorEncoder`

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,171 @@
...
+    double difference = goalPos - motorLencoder.getPosition();//RADIANS
+    if (Math.abs(difference) > encoderErrorTolerance){//even PID needs an acceptable error sometimes
+      //assuming calculate() is some sort of PID-esque thing
+      motor.setVoltage(armFeed.calculate(goalPos, FFvelocity, FFaccel)
+         + pid.calculate(motorLencoder.getPosition(), goalPos));
+      //motor.set(dir * armFeed.calculate(goalPos + Math.PI/2, FFvelocity, FFaccel));
+    } else {
+      //motor.set(dir * armFeed.calculate(goalPos + Math.PI/2, 0, 0));
+      motor.setVoltage(armFeed.calculate(goalPos, 0, 0)
+         + pid.calculate(motorLencoder.getPosition(), goalPos));
```
CoolSpy3: See thread in Mattermost

CoolSpy3: Consider using SparkMax onboard PIDController for faster adjustment

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,171 @@
...
+  public Arm(){
+    motorLencoder.setPositionConversionFactor(2*Math.PI/100);
+    motorLencoder.setPosition(-Math.PI / 2);
+    SmartDashboard.putNumber("FF: Velocity", FFvelocity);
+    SmartDashboard.putNumber("FF: Acceleration", FFaccel);
+    SmartDashboard.putNumber("GoalPosition", goalPos);
+    SmartDashboard.putNumber("KG", kG);
+    SmartDashboard.putString("Debooog", "No.");
+    SmartDashboard.putString("ArmENUM", 
+    snappedArmPos().toString()
```
CoolSpy3: tab over

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,171 @@
...
+
+  @Override
+  public void periodic() {
+    FFvelocity = SmartDashboard.getNumber("FF: Velocity", FFvelocity);
+    SmartDashboard.getNumber("KG", kG);
+    FFaccel = SmartDashboard.getNumber("FF: Acceleration", FFaccel);
+    goalPos = SmartDashboard.getNumber("GoalPosition", goalPos);
+    SmartDashboard.putNumber("ArmLencoderPos", motorLencoder.getPosition());
+    SmartDashboard.putString("ArmENUM", 
+    (closeSnappedArmPos() != null) ? closeSnappedArmPos().toString() : snappedArmPos().toString()
```
CoolSpy3: tab over

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
CoolSpy3: tab over

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,171 @@
+package org.carlmontrobotics.robotcode2023.subsystems;
```
CoolSpy3, line 1: Check that all curly brackets are preceded by a space

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
CoolSpy3: This logic can be made more clear by using a single setVoltage call and a ternary operator around the feedforward args

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
CoolSpy3: Velocity/Acceleration should change sign based on requested direction

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,171 @@
...
+  private double setpoint = 2.2;
+  private double FFvelocity = 6;
+  private double FFaccel = 6;
+  private ArmFeedforward armFeed = new ArmFeedforward(kS, kG, kV, kA);
+  private PIDController pid = new PIDController(Kp, Ki, Kd);
+  
+  public double goalPos;
+
+  public enum ArmPreset {
+    INTAKE(0), MID(1), HIGH(2);
```
CoolSpy3: Some of these values are out of range for the arm. They should be set to the correct values, and the range should be documented.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,171 @@
...
+  private double Kd = 1.2943;
+  private double Ki = 0.0;
+  
+  private double setpoint = 2.2;
+  private double FFvelocity = 6;
+  private double FFaccel = 6;
+  private ArmFeedforward armFeed = new ArmFeedforward(kS, kG, kV, kA);
+  private PIDController pid = new PIDController(Kp, Ki, Kd);
+  
+  public double goalPos;
```
CoolSpy3: Make private and add a getter

src/main/java/org/carlmontrobotics/robotcode2023/RobotContainer.java
```diff
@@ -4,25 +4,35 @@
...
 
 public class RobotContainer {
 
   public final Joystick driverController = new Joystick(0);
   public final Joystick manipulatorController = new Joystick(1);
   public final PowerDistribution pd = new PowerDistribution();
```
CoolSpy3, line 44: add newline

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,171 @@
...
+  /// these are all units ^ , actual arm0.15225 speed is determined by values in .calculate
+  private double Kp = 2.2985;
+  private double Kd = 1.2943;
+  private double Ki = 0.0;
+  
+  private double setpoint = 2.2;
+  private double FFvelocity = 6;
+  private double FFaccel = 6;
+  private ArmFeedforward armFeed = new ArmFeedforward(kS, kG, kV, kA);
+  private PIDController pid = new PIDController(Kp, Ki, Kd);
```
CoolSpy3: Make final

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
CoolSpy3: Use `SparkMaxAbsoluteEncoder`

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,171 @@
...
+        case HIGH: return MID;
+      }
+      return null;
+    }
+    
+  }
+
+  public Arm(){
+    motorLencoder.setPositionConversionFactor(2*Math.PI/100);
+    motorLencoder.setPosition(-Math.PI / 2);
```
CoolSpy3: After changing to an absolute encoder, use setZeroOffset with a value defined in constants

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
```diff
@@ -13,12 +15,23 @@
  * constants are needed, to reduce verbosity.
  */
 public final class Constants {
+
+    public static final class Arm{
```
CoolSpy3: Add space before `{`

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,171 @@
...
+import edu.wpi.first.math.controller.PIDController;
+import edu.wpi.first.math.util.Units;
+import edu.wpi.first.wpilibj2.command.SubsystemBase;
+import com.revrobotics.SparkMaxPIDController;
+import com.revrobotics.CANSparkMax;
+import com.revrobotics.RelativeEncoder;
+
+
+public class Arm extends SubsystemBase
+{
```
CoolSpy3: Move to previous line

CoolSpy3 (changes requested):
Here are my suggestions based on what I see rn. I recommend taking a look at these suggestions and then retesting to make sure the code still works.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
CoolSpy3: I thought about this, and actually, it makes more sense for our use case to have it as is. Please disregard this comment.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,101 @@
...
+    ArmPreset(double value) {
+      this.value = value;
+    }
+  }
+
+  public Arm() {
+    encoder.setPositionConversionFactor(1/15);
+    //encoder.setZeroOffset(-Math.PI / 2);
+    encoder.setPosition(-Math.PI / 2);
+    pid.setTolerance(2.5,10);
```
brettle: These numbers look wrong. Where did they come from?

brettle (changes requested):

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
FriedLongJohns: PLaceholder values.
Fixed now, though.

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
```diff
@@ -33,37 +160,101 @@ public static final class Roller {
...
+          public static int LT;
+          public static int RT;
+          public static int BACK;
+          public static int START;
 
-            public static final int rollerIntakeConeButton = Button.kA.value;
-            public static final int rollerOuttakeConeButton = Button.kB.value;
-            public static final int rollerIntakeCubeButton = Button.kX.value;
-            public static final int rollerOuttakeCubeButton = Button.kY.value;
-            public static final int rollerStopButton = Button.kRightBumper.value;
```
brettle, line 211: Why is this removed?

src/main/java/org/carlmontrobotics/robotcode2023/RobotContainer.java
```diff
@@ -23,25 +27,47 @@ public class RobotContainer {
...
+      .onTrue(new RunRoller(roller, RollerMode.INTAKE_CUBE, Constants.Roller.conePickupColor));
+    new JoystickButton(manipulatorController, Manipulator.intakeCone)
       .onTrue(new RunRoller(roller, RollerMode.INTAKE_CONE, Constants.Roller.conePickupColor));
-    new JoystickButton(manipulatorController, Manipulator.rollerIntakeCubeButton)
-      .onTrue(new RunRoller(roller, RollerMode.INTAKE_CUBE, Constants.Roller.cubePickupColor));
-    new JoystickButton(manipulatorController, Manipulator.rollerOuttakeConeButton)
-      .onFalse(new RunRoller(roller, RollerMode.OUTTAKE_CONE, Constants.Roller.conePickupColor));
-    new JoystickButton(manipulatorController, Manipulator.rollerOuttakeCubeButton)
-      .onFalse(new RunRoller(roller, RollerMode.OUTTAKE_CUBE, Constants.Roller.cubePickupColor));
-    new JoystickButton(manipulatorController, Manipulator.rollerStopButton).onTrue(new InstantCommand(() -> roller.setSpeed(0)));
```
brettle, line 87: Why is this removed?

src/main/java/org/carlmontrobotics/robotcode2023/RobotContainer.java
```diff
@@ -23,25 +27,47 @@ public class RobotContainer {
...
+  public final Arm arm = new Arm();
   public final Roller roller = new Roller();
 
   public RobotContainer() {
     configureButtonBindingsDriver();
     configureButtonBindingsManipulator();
+    arm.setDefaultCommand(new ArmPeriodic(
+      arm, 
+      () -> inputProcessing(getStickValue(manipulatorController, Axis.kLeftY)),
+      () -> inputProcessing(getStickValue(manipulatorController, Axis.kRightX))
```
brettle: Is that supposed to be an X?

src/main/java/org/carlmontrobotics/robotcode2023/RobotContainer.java
```diff
@@ -23,25 +27,47 @@ public class RobotContainer {
...
+    ));
   }
 
   private void configureButtonBindingsDriver() {}
 
   private void configureButtonBindingsManipulator() {
-    new JoystickButton(manipulatorController, Manipulator.rollerIntakeConeButton)
+    new JoystickButton(manipulatorController, Manipulator.toggleCube)
+      .onTrue(new InstantCommand(() -> {arm.object=0;}))
+      .onFalse(new InstantCommand(() -> {arm.object=1;}));
```
brettle: Use CONE and CUBE constants.

src/main/java/org/carlmontrobotics/robotcode2023/RobotContainer.java
```diff
@@ -23,25 +27,47 @@ public class RobotContainer {
...
+    new JoystickButton(manipulatorController, Manipulator.toggleFront)
+      .onTrue(new InstantCommand(() -> arm.isFront=true))
+      .onFalse(new InstantCommand(() -> arm.isFront=false));
```
brettle: I think these will cause the `onTrue` commands to run when the button is pressed and the `onFalse` commands to run when the button is released. So a pess-and-release of the button will not cause the values to toggle.

src/main/java/org/carlmontrobotics/robotcode2023/RobotContainer.java
```diff
@@ -23,25 +27,47 @@ public class RobotContainer {
...
+      new SetArmWristPosition(arm.getArmGoal(GoalPos.LOW), arm.getWristGoal(GoalPos.LOW), arm)
+    );
+    new JoystickButton(manipulatorController, Manipulator.mid).onTrue(
+      new SetArmWristPosition(arm.getArmGoal(GoalPos.MID), arm.getWristGoal(GoalPos.MID), arm)
+    );
+    new JoystickButton(manipulatorController, Manipulator.high).onTrue(
+      new SetArmWristPosition(arm.getArmGoal(GoalPos.HIGH), arm.getWristGoal(GoalPos.HIGH), arm)
+    );
+    new JoystickButton(manipulatorController, Manipulator.store).onTrue(
+      new SetArmWristPosition(arm.getArmGoal(GoalPos.STORED), arm.getWristGoal(GoalPos.STORED), arm)
```
brettle: Am I missing something or is this a duplicate of one of the earlier lines.

src/main/java/org/carlmontrobotics/robotcode2023/commands/ArmPeriodic.java
```diff
@@ -0,0 +1,111 @@
...
+    currArmRad = armSubsystem.getArmPos();
+    currWristRad = armSubsystem.getWristPos();
+    double currTime = Timer.getFPGATimestamp();
+    double deltaT = currTime - lastTime;
+    double[] goals = getRequestedSpeeds(deltaT);
+
+    TrapezoidProfile.Constraints armConstraints = new TrapezoidProfile.Constraints(MAX_FF_VEL[ARM], MAX_FF_ACCEL[ARM]);
+    var armProfile = new TrapezoidProfile(armConstraints,
+        new TrapezoidProfile.State(goals[0], goals[1]),
+        new TrapezoidProfile.State(currArmRad, armSubsystem.getArmPos()));
```
brettle: Isn't the second parameter supposed to be a velocity?

src/main/java/org/carlmontrobotics/robotcode2023/commands/ArmPeriodic.java
```diff
@@ -0,0 +1,111 @@
...
+
+    TrapezoidProfile.Constraints wristConstraints = new TrapezoidProfile.Constraints(MAX_FF_VEL[WRIST], MAX_FF_ACCEL[WRIST]);
+    var wristProfile = new TrapezoidProfile(wristConstraints,
+        new TrapezoidProfile.State(goals[2], goals[3]),
+        new TrapezoidProfile.State(currWristRad, armSubsystem.getWristPos()));
```
brettle: Isn't the second parameter supposed to be a velocity?

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,369 @@
...
+
+    public int object = CUBE;
+    public boolean isFront = true;
+
+    public Arm() {
+        armMotor.setInverted(inverted[ARM]);
+        wristMotor.setInverted(inverted[WRIST]);
+
+        armEncoder.setPositionConversionFactor(rotationToRad);
+        wristEncoder.setPositionConversionFactor(rotationToRad);
```
brettle, line 57: What about the velocity conversion factors?

src/main/java/org/carlmontrobotics/robotcode2023/commands/ArmPeriodic.java
```diff
@@ -0,0 +1,111 @@
...
+    wristSetpoint.position = armSubsystem.getWristClampedGoal(wristSetpoint.position);
+    armSubsystem.driveArm(armSetpoint);
+    armSubsystem.driveWrist(wristSetpoint);
+
+    if (Math.abs(goals[1]) < EPSILON)
+      armSubsystem.setArmTarget(armSubsystem.getArmPos());
+
+    if (Math.abs(goals[3]) < EPSILON) {
+      armSubsystem.setWristTarget(armSubsystem.getWristPos());
+    }
```
brettle: I don't think you want this. Imagine the arm is initially moving with a velocity of +1 rad/sec and the operator requests a speed of -1 rad/sec. Should something special happen when the speed is is temporarily near 0 along the way?

src/main/java/org/carlmontrobotics/robotcode2023/commands/ArmPeriodic.java
```diff
@@ -0,0 +1,111 @@
...
+    lastTime = currTime;
+  }
+
+  // Copy and pasted from drivetrain, handles input from joysticks
+  public double[] getRequestedSpeeds(double deltaT) {
```
brettle: Currently this method is misnamed because it returns more than just speeds. But instead of renaming the method, consider moving the position calculations into `execute()` and making this method just return speeds. That would also mean you wouldn't need to pass in `deltaT`.

src/main/java/org/carlmontrobotics/robotcode2023/commands/SetArmWristPosition.java
```diff
@@ -0,0 +1,48 @@
...
+public class SetArmWristPosition extends SequentialCommandGroup {
+
+    public SetArmWristPosition(double armPos, double wristPos, Arm arm) {
+        super();
+        // TODO: Test it!
+        //if arm is lower than MID when moving and if goal is lower than this by ANY AMOUNT, then fold the wrist first
+        //need separate flipped case for front too since values are flipped when arm passes through bellypan
+        if (
+                ( arm.isFront && (arm.getArmPos() < arm.getArmGoal(GoalPos.MID)) && (arm.getArmPos() > armPos)) ||
+                (!arm.isFront && (arm.getArmPos() > arm.getArmGoal(GoalPos.MID)) && (arm.getArmPos() < armPos))
```
brettle: It's not entirely clear what `isFront` is supposed to represent, but based on the code in `Arm` and the fact that the default value for `isFront` is `true`, I think it's supposed to mean we are intaking in the front and outtaking in the back (i.e. the more common situation), not that the arm itself is "in front" of the verticals. If you want to know whether the arm is "in front", you should calculate that based on the arm position.

Next, note that if `isFront` is true, `arm.getArmGoal(GoalPos.MID)` will return a position in the *back*. This code seems to be assuming that it would be in the front.

In general, I recommend against using `arm.getArmGoal(GoalPos.MID)` as the position for making these decisions since further testing could cause us to change the position in order to score effectively. It would be better to create a new constant representing the angular distance from vertical below which you need to tuck the wrist before making the arm more vertical.

src/main/java/org/carlmontrobotics/robotcode2023/commands/SetArmWristPosition.java
```diff
@@ -0,0 +1,48 @@
...
+             ){
+            //but have special case where arm is in front and wrist would collide, so we have to move arm pos to mid before folding
+            if (arm.isFront){//TODO MAKE SURE THIS LINE IS CORRECT
```
brettle: Read [this other comment where I suggest a rewrite](https://github.com/DeepBlueRobotics/RobotCode2023/pull/1#discussion_r1140978722) first.

This could happen on either side.

src/main/java/org/carlmontrobotics/robotcode2023/commands/SetArmWristPosition.java
```diff
@@ -0,0 +1,48 @@
...
+                super.addCommands(
+                    new InstantCommand(() -> arm.setArmTarget(GoalPos.MID[0].armPos)),
```
brettle: Read [this other comment where I suggest a rewrite](https://github.com/DeepBlueRobotics/RobotCode2023/pull/1#discussion_r1140978722) first.

This needs to be changed to move to a clearance position on the same side the arm is already on.

Also, if arm/wrist is currently in CAD's "Move" named position (or otherwise positioned such the end of the rollers is inside the robot below the height of the bumpers), moving the arm before stowing the wrist will result in a collision.

src/main/java/org/carlmontrobotics/robotcode2023/commands/SetArmWristPosition.java
```diff
@@ -0,0 +1,48 @@
...
+                    new WaitUntilCommand(arm.armPID::atSetpoint)
+                );
+            }
+            super.addCommands(
+                new InstantCommand(() -> arm.setWristTarget(wristStowPos)),
+                new WaitUntilCommand(arm.wristPID::atSetpoint),//always stow first, avoid collision
+                new InstantCommand(() -> arm.setArmTarget(armPos)),
+                new WaitUntilCommand(arm.armPID::atSetpoint),
+                new InstantCommand(() -> arm.setWristTarget(wristPos)),//assume wristPos isn't gonna collide with drivetrain because that'd be a stupid thing for people to do and I don't wanna add more if statements
```
brettle: Read [this other comment where I suggest a rewrite](https://github.com/DeepBlueRobotics/RobotCode2023/pull/1#discussion_r1140978722) first.

While I sympathize, people do stupid things all the time and we can't afford for them to result in a broken robot.

src/main/java/org/carlmontrobotics/robotcode2023/commands/SetArmWristPosition.java
```diff
@@ -0,0 +1,48 @@
...
+                new WaitUntilCommand(arm.wristPID::atSetpoint)
+            );
+        }else{
+            super.addCommands(//no collision? move both at once for moar speed
+                new InstantCommand(() -> arm.setArmTarget(armPos)),
+                new InstantCommand(() -> arm.setWristTarget(wristPos)),
+                new WaitUntilCommand(() -> {return (arm.wristPID.atSetpoint() && arm.armPID.atSetpoint());})
+            );
```
brettle: Read [this other comment where I suggest a rewrite](https://github.com/DeepBlueRobotics/RobotCode2023/pull/1#discussion_r1140978722) first.

This is a good plan *if* the arm is high enough to be clear *and* it is moving *higher* (i.e. away from vertical). However, if it is moving *lower* you could collide. Note that with your current logic, you will reach this code in that case.

src/main/java/org/carlmontrobotics/robotcode2023/commands/SetArmWristPosition.java
```diff
@@ -0,0 +1,48 @@
...
+import static org.carlmontrobotics.robotcode2023.Constants.GoalPos;
+import static org.carlmontrobotics.robotcode2023.Constants.Arm.wristStowPos;
+
+import org.carlmontrobotics.robotcode2023.subsystems.Arm;
+
+import edu.wpi.first.wpilibj2.command.InstantCommand;
+import edu.wpi.first.wpilibj2.command.SequentialCommandGroup;
+import edu.wpi.first.wpilibj2.command.WaitUntilCommand;
+
+public class SetArmWristPosition extends SequentialCommandGroup {
```
brettle, line 12: This probably needs a rewrite. Consider approaching the problem as follows:

1. Which of the following statements is true? "From any position there is always a direction I can rotate the arm joint without colliding" or "From any position there is always a direction I can rotate the wrist joint without colliding". Use the answer to your advantage by always rotating that joint in the available direction first.
2. To determine which direction to rotate it, draw a side view of the robot (including bumpers) and consider the various spots that the end of the roller could be. Can you find a relatively simple rule for determining which direction to rotate given where the end of the roller is?
3. Rotate the joint from step 1 in the direction from step 2 at least enough to allow you to rotate the other joint without colliding. Note that this doesn't necessarily mean that the first joint will always be rotated to the same position.
4. Rotate the second joint to a position that allows you full freedom in the rotation of the first joint.
5. Now to move to the target arm/wrist position from this position without colliding imagine you were *starting* at the target position followed steps 2-4 determine how to "get free". Do the resulting movements in reverse to get from your free position to the target position.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,369 @@
...
+    private final SparkMaxAbsoluteEncoder wristEncoder = wristMotor
+            .getAbsoluteEncoder(SparkMaxAbsoluteEncoder.Type.kDutyCycle);
+
+    private final SimpleMotorFeedforward armFeed = new SimpleMotorFeedforward(kS[ARM], kV[ARM], kA[ARM]);
+    private ArmFeedforward wristFeed = new ArmFeedforward(kS[WRIST], kG[WRIST], kV[WRIST], kA[WRIST]);
+
+    public PIDController armPID = new PIDController(kP[ARM], kI[ARM], kD[ARM]);
+    public PIDController wristPID = new PIDController(kP[WRIST], kI[WRIST], kD[WRIST]);
+
+    private SendableBuilder senb;
```
brettle: I don't think you need this.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,369 @@
...
+        wristEncoder.setPositionConversionFactor(rotationToRad);
+
+        wristEncoder.setZeroOffset(offsetRad[ARM]);
+        armEncoder.setZeroOffset(offsetRad[WRIST]);
+
+        armPID.setTolerance(posToleranceRad[ARM], velToleranceRadPSec[ARM]);
+        wristPID.setTolerance(posToleranceRad[WRIST], velToleranceRadPSec[WRIST]);
+
+        SmartDashboard.putBoolean("Toggle", false);
+        senb.update();
```
brettle: This should not be needed.

brettle: Instead, call `SmartDashboard.putData("Arm", this)`.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,369 @@
...
+        armEncoder.setPositionConversionFactor(rotationToRad);
+        wristEncoder.setPositionConversionFactor(rotationToRad);
+
+        wristEncoder.setZeroOffset(offsetRad[ARM]);
+        armEncoder.setZeroOffset(offsetRad[WRIST]);
+
+        armPID.setTolerance(posToleranceRad[ARM], velToleranceRadPSec[ARM]);
+        wristPID.setTolerance(posToleranceRad[WRIST], velToleranceRadPSec[WRIST]);
+
+        SmartDashboard.putBoolean("Toggle", false);
```
brettle: Not clear what this is for.

Note that you probably want an "Apply Risky Changes" button that, when clicked, sets the values that you don't want set with every keystroke in Shuffleboard. The setters for those properties would set an intermediate variable and the "Apply Risky Changes" would copy the values from those variables into the real variables.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,369 @@
...
+        // TODO: REMOVE THIS WHEN PID CONSTANTS ARE DONE
+        armPID.setP(kP[ARM]);
+        armPID.setI(kI[ARM]);
+        armPID.setD(kD[ARM]);
+        wristPID.setP(kP[WRIST]);
+        wristPID.setI(kI[WRIST]);
+        wristPID.setD(kD[WRIST]);
+        if (SmartDashboard.getBoolean("Toggle", false)) {
+            senb.update();
+        }
```
brettle: None of this should be needed.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,369 @@
...
+            wristFeedVolts = kgv;
+        }
+        // TODO: REMOVE WHEN DONE WITH TESTING (ANY CODE REVIEWERS, PLEASE REJECT MERGES
+        // TO MASTER IF THIS IS STILL HERE)
+        SmartDashboard.putNumber("WristFeedVolts", wristFeedVolts);
+        SmartDashboard.putNumber("WristPIDVolts", wristPIDVolts);
+        double volts = wristFeedVolts + wristPIDVolts;
+        SmartDashboard.putNumber("WristTotalVolts", volts);
+        wristMotor.setVoltage(volts);
+    }
```
brettle: These are no longer used.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,369 @@
...
+
+    public void driveArm(TrapezoidProfile.State profile) {
+        double kgv = getKg();
+        double armFeedVolts = kgv * getCoM().getAngle().getCos() + armFeed.calculate(profile.velocity, 0);
+        double armPIDVolts = armPID.calculate(armEncoder.getPosition(), profile.position);
```
brettle: `armEncoder.getPosition()` does not have the appropriate range.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,369 @@
...
+        SmartDashboard.putNumber("ArmPIDVolts", armPIDVolts);
+        double volts = armFeedVolts + armPIDVolts;
+        SmartDashboard.putNumber("ArmTotalVolts", volts);
+        armMotor.setVoltage(volts);
+    }
+
+    public void driveWrist(TrapezoidProfile.State profile) {
+        double kgv = wristFeed.calculate(getWristPosRelativeToGround(), 0, 0);
+        double wristFeedVolts = kgv * getCoM().getAngle().getCos() + wristFeed.calculate(profile.velocity, 0);
+        double wristPIDVolts = wristPID.calculate(wristEncoder.getPosition(), profile.position);
```
brettle: `wristEncoder.getPosition()` does not have the appropriate range.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,369 @@
...
+
+    @Override
+    public void initSendable(SendableBuilder builder) {
+        super.initSendable(builder);
+        senb=builder;
+        builder.addDoubleProperty("WristGoalPos", () -> goalPosRad[WRIST], goal -> goalPosRad[WRIST] = goal);
+        builder.addDoubleProperty("ArmGoalPos",   () -> goalPosRad[ARM], goal -> goalPosRad[ARM] = goal);
+        builder.addDoubleProperty("ArmPos",       () -> getArmPos(), null);
+        builder.addDoubleProperty("WisPos",       () -> getWristPos(), null);
+        builder.addDoubleProperty("RobotWisPos",  () -> getWristPosRelativeToGround(), null);
```
brettle: Wis -> Wrist

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,369 @@
...
+
+    public void toggleFront() {
+        isFront = !isFront;
+    }
+
+    public double getArmGoal(GoalPos[] pos) {
+        if (isFront) {
+            return pos[object].armPos;
+        } else {
+            return Rotation2d.fromRadians(-Math.PI).minus(Rotation2d.fromRadians(pos[object].armPos)).plus(Rotation2d.fromDegrees(10)).getRadians();
```
brettle: Positions on the opposite side can be unrelated to positions due to wrist asymmetry.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,369 @@
...
+        }
+    }
+
+    public double getWristGoal(GoalPos[] pos) {
+        if (isFront) {
+            return pos[object].wristPos;
+        } else {
+            return Rotation2d.fromRadians(pos[object].wristPos).unaryMinus().plus(Rotation2d.fromDegrees(10)).getRadians();
```
brettle: Positions on the opposite side can be unrelated to positions due to wrist asymmetry.

brettle (changes requested):

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
```diff
@@ -100,24 +99,23 @@ public static final class Arm {
     public static final class GoalPos {
 
         //#region Goal Positions
-        // TODO: PUT IN CORRECT POSITIONS
         // These positions are for if the intake/outtake takes place on the front (battery) side of the robot
-        // if intake/outtake on back, the "negative" pos will be used, which is offset by a a specific amount of degrees to account for wrist being in a different COM
+        // if intake/outtake on back, the "negative" pos will be used
         // 0 = CUBE, 1 = CONE
-        // {Cube, Cone}
+        // {Cone, Cube}
```
brettle: These 2 lines contradict each other.

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
```diff
@@ -132,9 +130,12 @@ public static final class GoalPos {
...
         };
         public static GoalPos[] INTAKE = {
-          new GoalPos(Units.degreesToRadians(-72.5), Units.degreesToRadians(73.37)),
-          new GoalPos(Units.degreesToRadians(-66.6), Units.degreesToRadians(29.94)) 
+          new GoalPos(Units.degreesToRadians(-72.5), Units.degreesToRadians(92.21)),
+          new GoalPos(Units.degreesToRadians(-66.6), Units.degreesToRadians(48.78)) 
         };
+        // Rollers are asymmetric, so when moving rollers to other side of robot, this value will correct
+        // the roller so that the rollers will act as if its on the front side
+        public static double asymmetricCorrectionDeg = 10;
```
brettle: Unfortunately, it's not that simple. Consider the case of placing a cube in the front instead of the back.

brettle (changes requested):

src/main/java/org/carlmontrobotics/robotcode2023/commands/ArmPeriodic.java
```diff
@@ -0,0 +1,105 @@
...
+
+  // Called when the command is initially scheduled.
+  @Override
+  public void initialize() {
+    lastTime = Timer.getFPGATimestamp();
+  }
+
+  // Called every time the scheduler runs while the command is scheduled.
+  @Override
+  public void execute() {
```
brettle: Some of what is in this method actually belongs in `Arm.periodic()`. As it stands, goals set by other commands will be ignored. Basically, what you want is for this method to just calculate the goal profile states based on the requested speeds, and then pass those goal profile states to the `Arm` subsystem.  Then `Arm.periodic()` should use whatever the the goal profile states are to create and calculate the profiles and drive the arm/wrist (i.e. the code which does that in this method should be moved to `Arm.periodic()`).

brettle (changes requested):

src/main/java/org/carlmontrobotics/robotcode2023/RobotContainer.java
ProfessorAtomicManiac: They are not meant to be toggle but a "whenHeld" button where if a button is held then the mode switches

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -194,17 +196,14 @@ public double getWristPosRelativeToGround() {
...
 
-    public void setArmTarget(double target) {
-        goalPosRad[ARM] = getArmClampedGoal(target);
-    }
-
-    public void setWristTarget(double target) {
-        goalPosRad[WRIST] = getWristClampedGoal(target);
+    public void setArmTarget(double targetPos, double targetVel) {
+        goalState[ARM].position = getArmClampedGoal(targetPos);
+        goalState[ARM].velocity = MathUtil.clamp(targetVel, 0, MAX_FF_VEL[ARM]);
```
brettle: This is wrong and not needed. Wrong because the goal velocity could be negative and not needed because the TrapezoidProfile velocity limit will take care of it.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -194,17 +196,14 @@ public double getWristPosRelativeToGround() {
...
     }
 
-    public void setArmWristTarget(double targetArm, double targetWrist) {
-        setArmTarget(targetArm);
-        setWristTarget(targetWrist);
+    public void setWristTarget(double targetPos, double targetVel) {
+        goalState[WRIST].position = getWristClampedGoal(targetPos);
+        goalState[WRIST].velocity = MathUtil.clamp(targetVel, 0, MAX_FF_VEL[WRIST]);
```
brettle: This is wrong and not needed. Wrong because the goal velocity could be negative and not needed because the TrapezoidProfile velocity limit will take care of it.

brettle (changes requested):

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,350 @@
...
+        // senb.update();
+        SmartDashboard.putData("Arm", this);
+        lastTime = Timer.getFPGATimestamp();
+
+    }
+
+    @Override
+    public void periodic() {
+        double currTime = Timer.getFPGATimestamp();
+        double deltaT = currTime - lastTime;
```
brettle: A bug is causing deltaT to keep growing. Do you see it?

brettle (changes requested):

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,350 @@
...
+        armMotor.setInverted(inverted[ARM]);
+        wristMotor.setInverted(inverted[WRIST]);
+
+        armEncoder.setPositionConversionFactor(rotationToRad);
+        wristEncoder.setPositionConversionFactor(rotationToRad);
+        armEncoder.setVelocityConversionFactor(rotationToRad);
+        wristEncoder.setVelocityConversionFactor(rotationToRad);
+
+        wristEncoder.setZeroOffset(offsetRad[ARM]);
+        armEncoder.setZeroOffset(offsetRad[WRIST]);
```
brettle: There are bugs on these 2 lines.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,350 @@
...
+        TrapezoidProfile.State armSetpoint = armProfile.calculate(deltaT);
+        TrapezoidProfile.State wristSetpoint = wristProfile.calculate(deltaT);
+        driveArm(armSetpoint);
+        driveWrist(wristSetpoint);
+    }
+
+    private void driveArm(TrapezoidProfile.State state) {
+        double kgv = getKg();
+        double armFeedVolts = kgv * getCoM().getAngle().getCos() + armFeed.calculate(state.velocity, 0);
+        double armPIDVolts = armPID.calculate(armEncoder.getPosition(), state.position);
```
brettle: armEncoder.getPosition() will not be in the correct range.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -0,0 +1,350 @@
...
+        armConstraints = new TrapezoidProfile.Constraints(MAX_FF_VEL[ARM], MAX_FF_ACCEL[ARM]);
+        armConstraints = new TrapezoidProfile.Constraints(MAX_FF_VEL[ARM], MAX_FF_ACCEL[ARM]);
+        var armProfile = new TrapezoidProfile(armConstraints,
+            goalState[ARM],
+            new TrapezoidProfile.State(getArmPos(), getArmVel()));
+    
+        wristConstraints = new TrapezoidProfile.Constraints(MAX_FF_VEL[WRIST], MAX_FF_ACCEL[WRIST]);
+        var wristProfile = new TrapezoidProfile(wristConstraints,
+            goalState[WRIST],
+            new TrapezoidProfile.State(getWristPos(), getWristVel()));
```
brettle: keep track of the current setpoints and create profiles
that start at the current setpoints (instead of the current positions).

src/main/java/org/carlmontrobotics/robotcode2023/commands/ArmPeriodic.java
```diff
@@ -0,0 +1,86 @@
...
+    double[] speeds = getRequestedSpeeds();
+    double currTime = Timer.getFPGATimestamp();
+    double deltaT = currTime - lastTime;
+
+    double goalArmRad = armSubsystem.getArmPos() + speeds[ARM] * deltaT;
+    double goalWristRad = armSubsystem.getWristPos() + speeds[WRIST] * deltaT;
```
brettle: Adjust the arm/wrist target positions based on the current target positions instead of current actual positions.

src/main/java/org/carlmontrobotics/robotcode2023/commands/ArmPeriodic.java
```diff
@@ -0,0 +1,86 @@
...
+
+    // Retrieve the profiled setpoint for the next timestep. This setpoint moves
+    // toward the goal while obeying the constraints.
+    goalArmRad = armSubsystem.getArmClampedGoal(goalArmRad);
+    goalWristRad = armSubsystem.getWristClampedGoal(goalWristRad);
```
brettle: These lines aren't needed because setArmTarget() is responsible for clamping.

src/main/java/org/carlmontrobotics/robotcode2023/commands/ArmPeriodic.java
```diff
@@ -0,0 +1,86 @@
...
+    armSubsystem.setArmTarget(goalArmRad, speeds[ARM]);
+    armSubsystem.setWristTarget(goalWristRad, speeds[WRIST]);
```
brettle: Don't update target target speeds. Just leave them at whatever the current target speeds are.

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
```diff
@@ -18,6 +21,177 @@
...
+        public static final double ARM_LOWER_LIMIT_RAD = -3 * Math.PI / 2;
+        public static final double ARM_UPPER_LIMIT_RAD = Math.PI / 2;
+        public static final double ARM_DISCONTINUITY_RAD = (ARM_LOWER_LIMIT_RAD + ARM_UPPER_LIMIT_RAD) / 2 - Math.PI;
+        public static final double WRIST_LOWER_LIMIT_RAD = -Math.PI;
+        public static final double WRIST_UPPER_LIMIT_RAD = Math.PI;
+        public static final double WRIST_DISCONTINUITY_RAD = (WRIST_LOWER_LIMIT_RAD + WRIST_UPPER_LIMIT_RAD) / 2 - Math.PI;
+
+        // TODO: Determine actual max vel/accel
+        public static double[] MAX_FF_VEL = {1, 1}; // rad / s
+        public static double[] MAX_FF_ACCEL = {1, 1}; // rad / s
```
brettle: Comment has wrong units.

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
```diff
@@ -18,6 +21,177 @@
...
+        public static final double[] kV = {4.3735, 1.6743};
+        public static final double[] kA = {0.24914, 0.032177};
+        public static final double kG_WRIST = .36214; // (V)
+
+        // PID
+        // FIXME BOTH WRIST AND ARM NEED TO TEST PID (Wrist PID never tested)
+        // Arm, Wrist
+        public static double[] kP = {6.7868, 4.7391}; // 4.2736 for arm from sysid was tested and it worked fine (V / rad)
+        public static double[] kI = {0, 0}; // (V / (rad * s) )
+        public static double[] kD = {4.4327, 0.69517}; // 0 for arm from sysid was tested and it worked fine (V / (rad / s) )
```
brettle: Why aren't we using the values from sysid? Were the new values arrived at via empirical tuning?

brettle (changes requested):

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
```diff
@@ -18,6 +21,177 @@
...
+
+        // Arm, Wrist
+        public static double[] posToleranceRad = { .05, .05 }; // rad
+        public static double[] velToleranceRadPSec = { 0.5, 0.5 }; // rad/s
+
+        // rad, rad/s
+        public static TrapezoidProfile.State[] goalState = { new TrapezoidProfile.State(-Math.PI / 2, 0), new TrapezoidProfile.State(0, 0) }; // rad
```
brettle: This isn't a constant. It should be a private member of Arm.java and very importantly it should be initialized to the current position at startup so that the arm/wrist don't try to move unless we tell them to.

brettle (approved):

PR #4, opened 2023-03-12
Drivetrain

CoolSpy3:


src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Drivetrain.java
```diff
@@ -0,0 +1,354 @@
...
+    @Override
+    public void periodic() {
+        for (SwerveModule module : modules) module.periodic();
+
+        // Update the odometry with current heading and encoder position
+        odometry.update(Rotation2d.fromDegrees(getHeading()), getModulePositions());
+
+        // Update the odometry with limelight
+        if(lime != null && lime.hasTarget()) {
+            odometry.addVisionMeasurement(lime.getTransform(Limelight.Transform.BOTPOSE).toPose2d(), WPIUtilJNI.now() * 1.0e-6 /* From the odometry.update implementation */);
```
brettle: The last parameter passed to `addVisionMeasurement()` here is incorrect. See here: https://docs.limelightvision.io/en/latest/apriltags_in_3d.html#using-wpilib-s-pose-estimator

brettle (changes requested):

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Drivetrain.java
CoolSpy3: My bad; I missed that! Should be fixed :D

src/main/deploy/pathplanner/135.path
ProfessorAtomicManiac, line 1: I heard that Uday worked on paths. Are these the exact files that he worked on?

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
```diff
@@ -13,9 +21,141 @@
...
+        public static final int canCoderPortFR = 2;
+        public static final int canCoderPortBL = 3;
+        public static final int canCoderPortBR = 4;
+
+        //#endregion
+
+        //#region Command Constants
+
+        public static final double kSlowDriveSpeed = 0.25;
+        public static final double kSlowDriveRotation = 0.30;
```
ProfessorAtomicManiac: add units for future reference?

src/main/deploy/pathplanner/135.path
CoolSpy3: I don't know, @Knight4 ?

CoolSpy3: You can also try pinging him on Mattermost if he doesn't respond here

src/main/java/org/carlmontrobotics/robotcode2023/commands/AlignChargingStation.java
```diff
@@ -0,0 +1,55 @@
...
+
+    @Override
+    public boolean isFinished() {
+        lastTime = (fwd ? Math.abs(getPitch()) < chargeStationAlignTolerance : Math.abs(getRoll()) < chargeStationAlignTolerance) ? lastTime == -1 ? System.currentTimeMillis() : lastTime : -1;
+        return System.currentTimeMillis() - lastTime > chargeStationAlignTime && lastTime != -1;
+    }
+
+    @Override
+    public void end(boolean interrupted) {
+        drivetrain.stop();
```
ProfessorAtomicManiac, line 53: When we were testing autopaths, if the wheels were not turned, the robot will slide when the charging station is unbalanced even in brake mode. We could potentially add something here to turn the wheels so the robot wouldn't slide.

src/main/java/org/carlmontrobotics/robotcode2023/commands/CorrectToPoint.java
```diff
@@ -0,0 +1,62 @@
...
+
+  // Called every time the scheduler runs while the command is scheduled.
+  @Override
+  public void execute() {
+    SmartDashboard.putNumber("Target X", setpoint.getX());
+    SmartDashboard.putNumber("Target Y", setpoint.getY());
+    SmartDashboard.putNumber("Target Theta", setpoint.getRotation().getDegrees());
+    dt.drive(pidX.calculate(dt.getPose().getX(), setpoint.getX()), 
+             pidY.calculate(dt.getPose().getY(), setpoint.getY()),
+             pidTheta.calculate(dt.getHeadingDeg(), setpoint.getRotation().getDegrees())); // TODO: make it turn -90 deg when told to turn 270 deg
```
ProfessorAtomicManiac: this. although easy fix would just to use "useContinuousInput" on pid controller

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Drivetrain.java
```diff
@@ -0,0 +1,355 @@
...
+        // SmartDashboard.putBoolean("Current Magnetic Field Disturbance",
+        // gyro.isMagneticDisturbance());
+
+        // drive to point testing
+
+        // PID testing for rotations
+        thetaPIDController[0] = SmartDashboard.getNumber("kpTheta", thetaPIDController[0]);
+        double temp = gyro.getAngle();
+        double tempTime = Timer.getFPGATimestamp();
+        SmartDashboard.putNumber("Angular Vel", Units.degreesToRadians(temp - prevAngle) / (tempTime - prevTime));
```
ProfessorAtomicManiac: you can remove this. This was just me testing whether the drivetrain can drive at a constant angular velocity (which it didn't). There probably is a method in the kinematics class that does this.

ProfessorAtomicManiac (commented):
I haven't really looked at ports since Uday has just posted them and he's missing some buttons, but overall I only commented on minor issues and its ready to be merged.

Webots/controllers/DeepBlueSim/DeepBlueSim.jar
brettle, line 1: Is this supposed to be committed or ignored?

src/main/java/org/carlmontrobotics/robotcode2023/commands/AlignChargingStation.java
```diff
@@ -0,0 +1,61 @@
...
+        fieldOriented = drivetrain.getFieldOriented();
+        drivetrain.setFieldOriented(false);
+    }
+
+    @Override
+    public void execute() {
+        double pitch = getPitch();
+        double roll = getRoll();
+        double forward = fwd && Math.abs(pitch) > chargeStationAlignTolerance ? chargeStationAlignSpeed * pitch + chargeStationAlignFF : 0;
+        double strafe = !fwd && Math.abs(roll) > chargeStationAlignTolerance ? chargeStationAlignSpeed * roll + chargeStationAlignFF : 0;
```
brettle: There is a bug on both of these lines.

src/main/java/org/carlmontrobotics/robotcode2023/Robot.java
```diff
@@ -19,17 +22,37 @@ public class Robot extends TimedRobot {
...
     CommandScheduler.getInstance().run();
     MotorErrors.printSparkMaxErrorMessages();
   }
 
   @Override
-  public void disabledInit() {}
+  public void disabledInit() {
+    new Thread(() -> {
+      try {
+        Thread.sleep(1000);
```
brettle: I'm not sure that is the best value this year given how scoring works.

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
```diff
@@ -13,9 +21,141 @@
...
+        // a = mu * 9.8 m/s^2
+        public static final double autoCentripetalAccel = mu * g * 2;
+
+        public static final boolean isGyroReversed = true;
+
+        // PID values are listed in the order kP, kI, and kD
+        public static final double[] xPIDController = {4, 0.0, 0.0};
+        public static final double[] yPIDController = {4, 0.0, 0.0};
+        public static final double[] thetaPIDController = {0.08, 0.0, 0.0};
+        public static final double[] tolerance = {1, 1, 5};
```
brettle: What are the units for tolerance? If it's meters, that seems rather large.

src/main/java/org/carlmontrobotics/robotcode2023/commands/RotateToFieldRelativeAngle.java
```diff
@@ -0,0 +1,43 @@
...
+        rotationPID.setSetpoint(angle.getDegrees());
+        rotationPID.setTolerance(THRESHOLD);
+        SendableRegistry.addChild(this, rotationPID);
+
+        addRequirements(drivetrain);
+    }
+
+    @Override
+    public void execute() {
+        drivetrain.drive(teleopDrive.currentForward, teleopDrive.currentStrafe, rotationPID.calculate(drivetrain.getHeading(), targetAngle.getDegrees()));
```
brettle: Doesn't this mean that the driver can't control the forward/strafe of the robot while this command is executing?

src/main/java/org/carlmontrobotics/robotcode2023/commands/RotateToFieldRelativeAngle.java
```diff
@@ -0,0 +1,43 @@
...
+    }
+    
+    @Override
+    public boolean isFinished() {
+        return rotationPID.atSetpoint();
```
brettle, line 38: What will happen if the robot still has angular velocity when it reaches the target angle?

src/main/java/org/carlmontrobotics/robotcode2023/commands/TeleopDrive.java
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  // Called every time the scheduler runs while the command is scheduled.
+  @Override
+  public void execute() {
+    double[] speeds = getRequestedSpeeds();
+    drivetrain.drive(speeds[0], speeds[1], speeds[2]);
+  }
+
+  public double[] getRequestedSpeeds() {double rawForward, rawStrafe, rotateClockwise, deltaT;
```
brettle: Fix formatting.

src/main/java/org/carlmontrobotics/robotcode2023/commands/TeleopDrive.java
```diff
@@ -0,0 +1,92 @@
...
+    deltaT = .05;
```
brettle: Where did this value come from?

src/main/java/org/carlmontrobotics/robotcode2023/commands/TeleopDrive.java
```diff
@@ -0,0 +1,92 @@
...
+    double targetAccelerationX = (rawForward - currentForward)/deltaT;
+    double targetAccelerationY = (rawStrafe - currentStrafe)/deltaT;
+    double accelerationMagnitude = Math.hypot(targetAccelerationX, targetAccelerationY);
+    if (accelerationMagnitude >= autoMaxAccelMps2) {
+      targetAccelerationX *= autoMaxAccelMps2/accelerationMagnitude;
+      targetAccelerationY *= autoMaxAccelMps2/accelerationMagnitude;
+    }
+    currentForward += targetAccelerationX*deltaT;
+    currentStrafe += targetAccelerationY*deltaT;
+    if (Math.abs(currentForward) <= Constants.OI.JOY_THRESH)
```
brettle: Units don't match.

src/main/java/org/carlmontrobotics/robotcode2023/commands/TeleopDrive.java
```diff
@@ -0,0 +1,92 @@
...
+      currentForward = 0;
+    if (Math.abs(currentStrafe) <= Constants.OI.JOY_THRESH)
```
brettle: Units don't match.

brettle (changes requested):

Webots/controllers/DeepBlueSim/DeepBlueSim.jar
CoolSpy3: It should be committed, otherwise the packaged `wbt` file points to a non-existent controller. In a separate note, I'm pretty sure that file is not up-to-date with the current robot, given that we haven't been using simulation, can we continue with this PR as-is, or should it be held until we can get around to updating the model?

src/main/java/org/carlmontrobotics/robotcode2023/commands/AlignChargingStation.java
CoolSpy3: Only bug I can see is that I'm pretty sure the "front" of the robot changed between when I wrote this code and now. I can't update that until I get a chance to see the new orientation of the gyro, but is there something else I missed?

src/main/java/org/carlmontrobotics/robotcode2023/Robot.java
CoolSpy3: Updated to 5000ms which should be above the 3s used for charge-station scoring

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
CoolSpy3: Units should be m/s, the values have been updated.

src/main/java/org/carlmontrobotics/robotcode2023/commands/RotateToFieldRelativeAngle.java
CoolSpy3: The idea is correct, the calls to teleopDrive should allow the driver to control forward and strafe. I'll agree that this usage, however, is incorrect, currentForward and currentStrafe are for teleopDrive internal use only. I've updated the code to use the correct method and privated those variables.

src/main/java/org/carlmontrobotics/robotcode2023/commands/RotateToFieldRelativeAngle.java
CoolSpy3: AFAIK, `atSetpoint` also checks the robot velocity, I've added code to set a velocity setpoint for this purpose using the same values as for `CorrectToPoint`.

src/main/java/org/carlmontrobotics/robotcode2023/commands/TeleopDrive.java
CoolSpy3: Idk, I've updated it to the robot period, we should retest to make sure the controls still have good gains.

CoolSpy3:
Comments should be addressed, I'll re-request reviews. If possible, I'd like to also retest teleop drive and charge station align to make sure nothing got broken.

Webots/controllers/DeepBlueSim/DeepBlueSim.jar
brettle: Ok. Don't let an old model hold up this PR.

src/main/java/org/carlmontrobotics/robotcode2023/commands/AlignChargingStation.java
brettle: Yes. Think about the range of values for `chargeStationAlignFF`.

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
```diff
@@ -129,17 +128,20 @@ public static final class Drivetrain {
...
+        public static final double kSlowDriveRotation = 0.30; // Percent Multiplier
         public static final double kAlignMultiplier = 1D/3D;
         public static final double kAlignForward = 0.6;
 
-        public static final double chargeStationAlignTolerance = 2.5;
+        public static final double chargeStationAlignTolerance = 2.5; // Degrees
         public static final double chargeStationAlignSpeed = 0.3 / 13;
         public static final double chargeStationAlignTime = 500;
         public static final double chargeStationAlignFF = 0.1;
+        public static final double wheelTurnDriveSpeed = 0.0001; // Meters / Second ; A non-zero speed just used to orient the wheels to the correct angle. This should be very small to avoid actually moving the robot.
```
brettle, line 142: The wheels can just spin in place. I don't think this should be needed, but I don't object.

brettle (dismissed):

brettle (changes requested):
chargeStationAlignFF issue needs to be addressed.

src/main/java/org/carlmontrobotics/robotcode2023/commands/AlignChargingStation.java
CoolSpy3: Ah....

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
CoolSpy3: Agree that it's not needed, but `drive` already handles all the necessary kinematics/wheel optimization/etc. Better is the enemy of good enough (especially when we don't have to maintain 2 drive pipelines). In this case I think it should be fine. The code calls `dt.stop()` immediately after, so drive speed should (ultimately) get set to zero. This is just to get around the "don't move if speed == 0" line.

CoolSpy3:
Alright, at this point, I think the code should be good for review. Even if it passes, I still want to conduct the aforementioned tests before this PR is merged.

src/main/java/org/carlmontrobotics/robotcode2023/commands/AlignChargingStation.java
```diff
@@ -0,0 +1,64 @@
...
+        fieldOriented = drivetrain.getFieldOriented();
+        drivetrain.setFieldOriented(false);
+    }
+
+    @Override
+    public void execute() {
+        double pitch = getPitch();
+        double roll = getRoll();
+        double forward = fwd && Math.abs(pitch) > chargeStationAlignTolerance ? chargeStationAlignSpeed * pitch + Math.copySign(chargeStationAlignFF, pitch) : 0;
+        double strafe = !fwd && Math.abs(roll) > chargeStationAlignTolerance ? chargeStationAlignSpeed * roll + Math.copySign(chargeStationAlignFF, roll) : 0;
```
brettle: This might "work" but at a minimum `chargeStationAlignSpeed` and `chargeStationAlignFF` should be renamed because they don't have their normal units of m/s and volts respectively.

Even better, use the drivetrain's `SimpleFeedForward` model to compute the voltage required to overcome the tilt and move at the desired velocity.

brettle (changes requested):

src/main/java/org/carlmontrobotics/robotcode2023/commands/AlignChargingStation.java
CoolSpy3: I've updated the variable names to include units. As for using a Feedforward model, atm I don't think we have FF constants for the drivetrain as a whole, just the individual wheels. Also, those don't account for gravity, instead assuming that it will always be canceled out by the normal force.

CoolSpy3:
My bad, review requested on wrong PR.

brettle (approved):

src/main/java/org/carlmontrobotics/robotcode2023/commands/AlignChargingStation.java
brettle: To keep things simple I'm fine with the current approach for now.

Fwiw, if you wanted to test what I was thinking of, you'd set the `pitchSupplier` and `rollSupplier` passed to the `SwerveModule` constructor to return the gyro's current pitch and roll (instead of just the initial pitch and roll they currently return) and set `chargeStationAlignSpeedMpSPerDeg` to 0. Basically `SwerveModule` already has the anti-gravity code. We're just not using it at the moment.

CoolSpy3: Ah, yes I remember proposing that in the initial versions of the command in addition to what we have now. Can't recall why we ended up with this solution. We should maybe consider it for future revisions, but atm I'm inclined to keep it as-is.

CoolSpy3:
Marking this PR as a draft until aforementioned testing is complete...

ProfessorAtomicManiac (approved):

PR #6, opened 2023-03-16
Merging ArmTest into Arm branch

ProfessorAtomicManiac:


ProfessorAtomicManiac:
I'm setting this pull request as "Ready for Review", but ignore `SetArmWristPosition.java` and the `GoalPos` in `Constants.java` until either Alex and Sofie push their commits (or I do them if they don't get it in later tonight). There is a lot to review anyways even excluding those files.

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
```diff
@@ -59,49 +66,31 @@ public static final class Arm {
...
         // how far it is from the min/max until it should stop the arm so that arm doesn't break robot
-        public static final double ARM_ANGLE_TOLERANCE = Math.PI / 2;
-        public static final double WRIST_ANGLE_MIN = -Math.PI;
-        public static final double WRIST_ANGLE_MAX = Math.PI;
+        public static final double ARM_ANGLE_TOLERANCE_RAD = Math.PI / 4;
+        public static final double WRIST_ANGLE_MIN_RAD = -Math.PI;
+        public static final double WRIST_ANGLE_MAX_RAD = Math.PI;
         // how far it is from the min/max until it should stop the wrist so that wrist doesn't break robot
-        public static final double WRIST_ANGLE_TOLERANCE = Math.PI / 6;
+        public static final double WRIST_ANGLE_TOLERANCE_RAD = Math.PI / 6;
```
brettle: While the above works (mostly), I have several concerns:
1. Despite the comments, I had to look at exactly how these values were being used to understand the intended meaning of each one.
2. If `*_ANGLE_MAX_RAD-*ANGLE_MIN_RAD != 2*Math.PI` bad things will happen.
3. The "tolerance" is assumed to be symmetric, which isn't the case for the wrist.
4. The important measurable values (i.e. the actual upper and lower limits of the joints) are not specified directly. They are only specified indirectly (i.e. by combining the various values).

I recommend instead just specifying `*_LOWER_LIMIT_RAD` and `*_UPPER_LIMIT_RAD` to correspond to the desired soft stops (which should come from design), with the only restriction being that `*_UPPER_LIMIT_RAD > *_LOWER_LIMIT_RAD`. Then *compute* where the discontinuity should be as halfway between those angles (i.e. `*_DISCONTINUITY_RAD = (*_LOWER_LIMIT_RAD + *_UPPER_LIMIT_RAD) / 2 - Math.PI`). Then to find the modulus of angle theta, call `MathUtil.inputModulus(theta,  `*_DISCONTINUITY_RAD` *_DISCONTINUITY_RAD + 2 * Math.PI)`.

src/main/java/org/carlmontrobotics/robotcode2023/commands/ArmPeriodic.java
```diff
@@ -29,47 +29,67 @@ public ArmPeriodic(Arm armSubsystem, DoubleSupplier arm, DoubleSupplier wrist) {
...
   // Called every time the scheduler runs while the command is scheduled.
   @Override
   public void execute() {
     double[] speeds = getRequestedSpeeds();
-    armSubsystem.driveArm(speeds[0], 0);
-    armSubsystem.driveWrist(speeds[1], 0);
-    armSubsystem.setArmTarget(armSubsystem.getArmPos());
-    armSubsystem.setWristTarget(armSubsystem.getWristPos());
+    // if driven by joysticks, run only feedforward
+    if (speeds[0] == 0)
```
brettle:
```suggestion
    if (speeds[ARM] == 0) 
```

src/main/java/org/carlmontrobotics/robotcode2023/commands/ArmPeriodic.java
```diff
@@ -29,47 +29,67 @@ public ArmPeriodic(Arm armSubsystem, DoubleSupplier arm, DoubleSupplier wrist) {
...
-    armSubsystem.setArmTarget(armSubsystem.getArmPos());
-    armSubsystem.setWristTarget(armSubsystem.getWristPos());
+    // if driven by joysticks, run only feedforward
+    if (speeds[0] == 0) 
+      armSubsystem.driveArm(0, 0, true);
+    else {
+      armSubsystem.driveArm(speeds[ARM], 0, false);
+      armSubsystem.setArmTarget(armSubsystem.getArmPos());
+    }
+    if (speeds[1] == 0)
```
brettle:
```suggestion
    if (speeds[WRIST] == 0)
```

src/main/java/org/carlmontrobotics/robotcode2023/commands/ArmPeriodic.java
```diff
@@ -29,47 +29,67 @@ public ArmPeriodic(Arm armSubsystem, DoubleSupplier arm, DoubleSupplier wrist) {
...
+      armSubsystem.driveWrist(0, 0, true);
+    else {
+      armSubsystem.driveWrist(speeds[WRIST], 0, false);
+      armSubsystem.setWristTarget(armSubsystem.getWristPos());
+    }
```
brettle: I recommend that you replace all of this with something that just updates the target positions based on the current positions, the requested speeds, and the timestep. Then let the `Arm` subsystem worry about actually getting to the new desired position.

src/main/java/org/carlmontrobotics/robotcode2023/commands/ArmPeriodic.java
```diff
@@ -29,47 +29,67 @@ public ArmPeriodic(Arm armSubsystem, DoubleSupplier arm, DoubleSupplier wrist) {
...
+    double targetAccelerationArm = (rawArm - currentArm) / deltaT;
+    double targetAccelerationWrist = (rawWrist - currentWrist) / deltaT;
+    if (targetAccelerationArm >= MAX_FF_ACCEL[ARM]) {
+      targetAccelerationArm *= MAX_FF_ACCEL[ARM] / targetAccelerationArm;
     }
-    if (targetAccelerationWrist >= MAX_FF_ACCEL) {
-      targetAccelerationWrist *= MAX_FF_ACCEL/targetAccelerationWrist;
+    if (targetAccelerationWrist >= MAX_FF_ACCEL[WRIST]) {
+      targetAccelerationWrist *= MAX_FF_ACCEL[WRIST] / targetAccelerationWrist;
     }
```
brettle: Unless I'm missing something, the only difference between these lines a couple calls to `Math.max()` are that these lines can introduce divide-by-zero errors.

brettle: The current implementation also only limits the acceleration in one direction.

src/main/java/org/carlmontrobotics/robotcode2023/commands/ArmPeriodic.java
```diff
@@ -29,47 +29,67 @@ public ArmPeriodic(Arm armSubsystem, DoubleSupplier arm, DoubleSupplier wrist) {
...
-    currentArm += targetAccelerationArm*deltaT;
-    currentWrist += targetAccelerationWrist*deltaT;
+    currentArm += targetAccelerationArm * deltaT;
+    currentWrist += targetAccelerationWrist * deltaT;
     if (Math.abs(currentArm) <= Constants.OI.JOY_THRESH)
       currentArm = 0;
     if (Math.abs(currentWrist) <= Constants.OI.JOY_THRESH)
       currentWrist = 0;
```
brettle: Units mismatch.

src/main/java/org/carlmontrobotics/robotcode2023/commands/ArmPeriodic.java
brettle: Even better, pass the Arm subsystem a goal which is a `TrapezoidalProfile.State` containing the desired position and the requested speed and have the Arm subsystem execute a motion profile to achieve that state.

src/main/java/org/carlmontrobotics/robotcode2023/commands/ArmPeriodic.java
```diff
@@ -29,47 +29,67 @@ public ArmPeriodic(Arm armSubsystem, DoubleSupplier arm, DoubleSupplier wrist) {
...
     }
-    if (targetAccelerationWrist >= MAX_FF_ACCEL) {
-      targetAccelerationWrist *= MAX_FF_ACCEL/targetAccelerationWrist;
+    if (targetAccelerationWrist >= MAX_FF_ACCEL[WRIST]) {
+      targetAccelerationWrist *= MAX_FF_ACCEL[WRIST] / targetAccelerationWrist;
     }
-    currentArm += targetAccelerationArm*deltaT;
-    currentWrist += targetAccelerationWrist*deltaT;
+    currentArm += targetAccelerationArm * deltaT;
+    currentWrist += targetAccelerationWrist * deltaT;
```
brettle: Limiting acceleration is probably best achieved by creating a motion profile in `Arm`.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -15,148 +15,110 @@
...
     private final SimpleMotorFeedforward armFeed = new SimpleMotorFeedforward(kS[ARM], kV[ARM], kA[ARM]);
-    private final ArmFeedforward wristFeed = new ArmFeedforward(kS[WRIST], kG_WRIST, kV[WRIST], kA[WRIST]);
+    private ArmFeedforward wristFeed = new ArmFeedforward(kS[WRIST], kG[WRIST], kV[WRIST], kA[WRIST]);
 
-    private final PIDController armPID = new PIDController(kP[ARM], kI[ARM], kD[ARM]);
-    private final PIDController wristPID = new PIDController(kP[WRIST], kI[WRIST], kD[WRIST]);
-    // 0 = cube, 1 = cone (used ints to make array indexing easier)
-    private int isCube = 0;
+    private PIDController armPID = new PIDController(kP[0], kI[0], kD[0]);
+    private PIDController wristPID = new PIDController(kP[1], kI[1], kD[1]);
```
brettle: You stopped using `ARM` and `WRIST` and indices.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -15,148 +15,110 @@
...
-        SmartDashboard.putNumber("IntakeArmCone", GoalPos.INTAKE[WRIST].armPos);
-        SmartDashboard.putNumber("IntakeWristCone", GoalPos.INTAKE[WRIST].wristPos);
-
-        SmartDashboard.putNumber("Max FF Vel", MAX_FF_VEL);
-        SmartDashboard.putNumber("Max FF Accel", MAX_FF_ACCEL);
+        putArmControlOnSmartDashboard();
     }
 
     @Override
     public void periodic() {
```
brettle, line 63: This method should actually move the arm/wrist toward the goal position/velocity (preferably via a motion profile). Right now, it doesn't actually move the arm/wrist at all. In fact, as far as I can tell, nothing in the code would actually move the arm/wrist, because nothing calls `setArmVoltage` or `setWristVoltage`.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -15,148 +15,110 @@
...
+            SmartDashboard.putNumber("WristFeedVolts", wristFeedVolts);
+            SmartDashboard.putNumber("WristPIDVolts", wristPIDVolts);
+            return (usePID ? wristFeedVolts + wristPIDVolts : wristFeedVolts);
+        }
+        double wristFeedVolts = wristFeed.calculate(getWristPosRelativeToGround(), vel, accel);
+        double wristPIDVolts = wristPID.calculate(getWristPos(), goalPosRad[WRIST]);
+        SmartDashboard.putNumber("WristFeedVolts", wristFeedVolts);
+        SmartDashboard.putNumber("WristPIDVolts", wristPIDVolts);
+        return (usePID ? wristFeedVolts + wristPIDVolts : wristFeedVolts);
     }
```
brettle: In each of these 2 methods, the only difference between what is in the if and what isn't is that vel and accel are 0. Please don't duplicate code unnecessarily.

brettle (changes requested):
I have more, but this is all I have time for now.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -15,148 +15,110 @@
...
+        getArmControlOnSmartDashboard();
     }
 
     public void setArmVoltage(double volts) {
         armMotor.setVoltage(volts);
     }
+
     public void setWristVoltage(double volts) {
         wristMotor.setVoltage(volts);
     }
```
brettle: These probably don't need to be public.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -15,148 +15,110 @@
...
-    public double driveArm(double vel, double accel) {
-        if (Math.abs(ARM_ANGLE_MAX - getArmPos()) < ARM_ANGLE_TOLERANCE)
-            return getKg() * getCoM().getAngle().getCos() + armFeed.calculate(0, 0) + armPID.calculate(armEncoder.getPosition(), goalPosRad[ARM]);
-        return getKg() * getCoM().getAngle().getCos() + armFeed.calculate(vel, accel) + armPID.calculate(armEncoder.getPosition(), goalPosRad[ARM]);
+
+    public double driveArm(double vel, double accel, boolean usePID) {
```
brettle: Probably doesn't need to be public and should be renamed to something like `computeArmVoltage()` since it doesn't actually drive anything.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -15,148 +15,110 @@
...
+        SmartDashboard.putNumber("ArmFeedVolts", armFeedVolts);
+        SmartDashboard.putNumber("ArmPIDVolts", armPIDVolts);
+        return (usePID ? armFeedVolts + armPIDVolts : armFeedVolts);
     }
-    public double driveWrist(double vel, double accel) {
-        if (Math.abs(WRIST_ANGLE_MAX - getWristPos()) < WRIST_ANGLE_TOLERANCE)
-            return wristFeed.calculate(getWristPosRelativeToGround(), 0, 0) + wristPID.calculate(getWristPos(), goalPosRad[WRIST]);
-        return wristFeed.calculate(getWristPosRelativeToGround(), vel, accel) + wristPID.calculate(getWristPos(), goalPosRad[WRIST]);
+
+    public double driveWrist(double vel, double accel, boolean usePID) {
```
brettle: Probably doesn't need to be public and should be renamed to something like `computeWristVoltage()` since it doesn't actually drive anything.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -15,148 +15,110 @@
...
+        if (Math.abs(WRIST_ANGLE_MAX_RAD - getWristPos()) < WRIST_ANGLE_TOLERANCE_RAD || Math.abs(WRIST_ANGLE_MIN_RAD - getArmPos()) < ARM_ANGLE_TOLERANCE_RAD) {
```
brettle: There is a bug on this line.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -202,46 +168,157 @@ public boolean wristAtSetpoint() {
...
+
     public double getWristGoal(GoalPos[] pos) {
         if (isFront) {
-            return pos[isCube].wristPos;
+            return pos[object].wristPos;
         } else {
-            return Rotation2d.fromRadians(pos[isCube].wristPos).unaryMinus().getRadians();
+            return Rotation2d.fromRadians(pos[object].wristPos).unaryMinus().getRadians();
         }
     }
```
brettle: The asymmetry of the roller means that you can't just reflect the relevant angles across the y-axis. You'll need an entirely different set of positions for the opposite side.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -202,46 +168,157 @@ public boolean wristAtSetpoint() {
...
+    }
+
+    public void getPIDOnSmartDashboard() {
+        kP[WRIST] = SmartDashboard.getNumber("kP: Wis", kP[WRIST]);
+        kI[WRIST] = SmartDashboard.getNumber("kI: Wis", kI[WRIST]);
+        kD[WRIST] = SmartDashboard.getNumber("kD: Wis", kD[WRIST]);
+        kP[ARM] = SmartDashboard.getNumber("kP: Arm", kP[ARM]);
+        kI[ARM] = SmartDashboard.getNumber("kI: Arm", kI[ARM]);
+        kD[ARM] = SmartDashboard.getNumber("kD: Arm", kD[ARM]);
+    }
```
brettle: Fwiw, these methods are never called.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -202,46 +168,157 @@ public boolean wristAtSetpoint() {
...
+        MAX_FF_ACCEL[ARM] = SmartDashboard.getNumber("Max FF Accel Arm", MAX_FF_ACCEL[ARM]);
+    }
+
+    @Override
+    public void initSendable(SendableBuilder builder) {
+        super.initSendable(builder);
+        builder.addDoubleProperty("WristGoalPos", () -> goalPosRad[1], goal -> goalPosRad[1] = goal);
+        builder.addDoubleProperty("ArmPos", () -> getArmPos(), null);
+        builder.addDoubleProperty("WisPos", () -> getWristPos(), null);
+        builder.addDoubleProperty("RobotWisPos", () -> getWristPosRelativeToGround(), null);
```
brettle: Typos in names.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -202,46 +168,157 @@ public boolean wristAtSetpoint() {
...
+        builder.addDoubleProperty("kP: Arm", () -> kP[ARM], x -> kP[ARM] = x);
+        builder.addDoubleProperty("kP: Wis", () -> kP[WRIST], x -> kP[WRIST] = x);
+        builder.addDoubleProperty("kI: Arm", () -> kI[ARM], x -> kI[ARM] = x);
+        builder.addDoubleProperty("kI: Wis", () -> kI[WRIST], x -> kI[WRIST] = x);
+        builder.addDoubleProperty("kD: Arm", () -> kD[ARM], x -> kD[ARM] = x);
+        builder.addDoubleProperty("kD: Wis", () -> kD[WRIST], x -> kD[WRIST] = x);
```
brettle, line 238: Consider replacing your `{get,put}*OnSmartDashboard()` methods with additional lines here.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Roller.java
```diff
@@ -101,27 +82,33 @@ public boolean hasGamePiece() {
...
+    public void getRollerConstsOnSmartDashboard() {
+        RollerMode.INTAKE_CONE.speed = SmartDashboard.getNumber("Intake Cone Speed", RollerMode.INTAKE_CONE.speed);
+        RollerMode.OUTTAKE_CONE.speed = SmartDashboard.getNumber("Outtake Cone Speed", RollerMode.OUTTAKE_CONE.speed);
+        RollerMode.INTAKE_CUBE.speed = SmartDashboard.getNumber("Intake Cube Speed", RollerMode.INTAKE_CUBE.speed);
+        RollerMode.OUTTAKE_CUBE.speed = SmartDashboard.getNumber("Outtake Cube Speed", RollerMode.OUTTAKE_CUBE.speed);
+        RollerMode.INTAKE_CONE.time = SmartDashboard.getNumber("Intake Cone Time", RollerMode.INTAKE_CONE.time);
+        RollerMode.OUTTAKE_CONE.time = SmartDashboard.getNumber("Outtake Cone Time", RollerMode.OUTTAKE_CONE.time);
+        RollerMode.INTAKE_CUBE.time = SmartDashboard.getNumber("Intake Cube Time", RollerMode.INTAKE_CUBE.time);
+        RollerMode.OUTTAKE_CUBE.time = SmartDashboard.getNumber("Outtake Cube Time", RollerMode.OUTTAKE_CUBE.time);
     }
```
brettle, line 113: Consider moving this functionality into an `initSendable()` method.

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
```diff
@@ -23,27 +23,34 @@ public static final class Arm {
...
+        public static final int CUBE = 0;
+        public static final int CONE = 1;
+
         // Feedforward
-        // FIXME WRIST NEEDS SYSID DONE
         // Arm, Wrist
-        public static final double[] kS = { .067766, .074798 }; // (V)
-        public static final double[] kV = { .019762, 1.6743 }; // (V / rad/s)
-        public static final double[] kA = { .00039212, 0.032177 }; // (V / rad/s^2)
+        // kS, kG, kV, kA
```
CoolSpy3:
```suggestion
```
Given the format of the following variables, this line is unnecessary.

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
```diff
@@ -59,49 +66,30 @@ public static final class Arm {
...
-        public static final double WRIST_ANGLE_MAX = Math.PI;
-        // how far it is from the min/max until it should stop the wrist so that wrist doesn't break robot
-        public static final double WRIST_ANGLE_TOLERANCE = Math.PI / 6;
-
-        public static double MAX_FF_VEL = 0.1; // rad / s
-        public static double MAX_FF_ACCEL = 0.11; // rad / s
+        // TODO: Replace these values with Design's actual values
+        public static final double ARM_LOWER_LIMIT_RAD = -3 * Math.PI / 2;
+        public static final double ARM_UPPER_LIMIT_RAD = Math.PI / 2;
+        public static final double ARM_DISCONTINUITY_RAD = (ARM_LOWER_LIMIT_RAD + ARM_UPPER_LIMIT_RAD) / 2 - Math.PI;
```
CoolSpy3, line 73: Can you bound this from 0 to 2pi?

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
```diff
@@ -59,49 +66,30 @@ public static final class Arm {
...
+        public static final double WRIST_LOWER_LIMIT_RAD = -Math.PI;
+        public static final double WRIST_UPPER_LIMIT_RAD = Math.PI;
+        public static final double WRIST_DISCONTINUITY_RAD = (WRIST_LOWER_LIMIT_RAD + WRIST_UPPER_LIMIT_RAD) / 2 - Math.PI;
```
CoolSpy3, line 76: Can you bound this from 0 to 2pi?

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
```diff
@@ -138,19 +126,34 @@ public static final class Roller {
...
         //#region Command Constants
-
-        public static final double coneIntakeConeOuttakeSpeed = -.3;
-        public static final double coneOuttakeConeIntakeSpeed = .7;
+        // TODO: Determine actual speeds/timings for roller
+        public static class RollerMode {
+            public static RollerMode INTAKE_CONE = new RollerMode(-0.5, .5, true);
+            public static RollerMode INTAKE_CUBE = new RollerMode(0.3, .25, true);
+            public static RollerMode OUTTAKE_CONE = new RollerMode(0.5, .5, false);
+            public static RollerMode OUTTAKE_CUBE = new RollerMode(-0.5, .5, false);
```
CoolSpy3, line 175: Make `final`

src/main/java/org/carlmontrobotics/robotcode2023/commands/RunRoller.java
```diff
@@ -2,13 +2,15 @@
...
+import org.carlmontrobotics.robotcode2023.Constants.Roller.RollerMode;
 import org.carlmontrobotics.robotcode2023.subsystems.Roller;
-import org.carlmontrobotics.robotcode2023.subsystems.Roller.RollerMode;
+
 
 import edu.wpi.first.wpilibj.Timer;
 import edu.wpi.first.wpilibj.smartdashboard.SmartDashboard;
 import edu.wpi.first.wpilibj2.command.CommandBase;
 
+
```
CoolSpy3:
```suggestion
```

src/main/java/org/carlmontrobotics/robotcode2023/commands/RunRoller.java
```diff
@@ -2,13 +2,15 @@
 
 import java.awt.Color;
 
+import org.carlmontrobotics.robotcode2023.Constants.Roller.RollerMode;
 import org.carlmontrobotics.robotcode2023.subsystems.Roller;
-import org.carlmontrobotics.robotcode2023.subsystems.Roller.RollerMode;
+
```
CoolSpy3, line 7:
```suggestion
```

brettle (changes requested):

src/main/java/org/carlmontrobotics/robotcode2023/commands/SetArmWristPosition.java
```diff
@@ -17,6 +16,7 @@ public class SetArmWristPosition extends ConditionalCommand {
     private final static double armSideTransferOffsetFromVertical = Units.degreesToRadians(45);
 
     public SetArmWristPosition(double armPos, double wristPos, Arm arm) {
+        // TODO: Alex simplify this command
```
CoolSpy3, line 14: TODO

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Roller.java
```diff
@@ -101,27 +82,33 @@ public boolean hasGamePiece() {
...
 
     public double getGamePieceDistanceIn() {
-        return Units.metersToInches((distSensor.getRange() - distSensorDepthMM /* The sensor measures from the back of the sensor */) /1000 /* Convert mm to m */);
+        return Units.metersToInches((distSensor.getRange() - distSensorDepthMM /*
+                                                                                * The sensor measures from the back of
+                                                                                * the sensor
+                                                                                */) / 1000 /* Convert mm to m */);
```
CoolSpy3, line 88: Please undo this formatting change or adjust it so that individual comments do not span lines and there's not 18 tabs

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -15,148 +15,98 @@
...
     private final SimpleMotorFeedforward armFeed = new SimpleMotorFeedforward(kS[ARM], kV[ARM], kA[ARM]);
-    private final ArmFeedforward wristFeed = new ArmFeedforward(kS[WRIST], kG_WRIST, kV[WRIST], kA[WRIST]);
+    private ArmFeedforward wristFeed = new ArmFeedforward(kS[WRIST], kG[WRIST], kV[WRIST], kA[WRIST]);
 
-    private final PIDController armPID = new PIDController(kP[ARM], kI[ARM], kD[ARM]);
-    private final PIDController wristPID = new PIDController(kP[WRIST], kI[WRIST], kD[WRIST]);
-    // 0 = cube, 1 = cone (used ints to make array indexing easier)
-    private int isCube = 0;
+    private PIDController armPID = new PIDController(kP[ARM], kI[ARM], kD[ARM]);
+    private PIDController wristPID = new PIDController(kP[WRIST], kI[WRIST], kD[WRIST]);
```
CoolSpy3: Make `final`

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -15,148 +15,98 @@
...
-        SmartDashboard.putNumber("Max FF Vel", MAX_FF_VEL);
-        SmartDashboard.putNumber("Max FF Accel", MAX_FF_ACCEL);
+        putArmControlOnSmartDashboard();
     }
 
     @Override
     public void periodic() {
-        
-        //TODO: REMOVE WHEN DONE WITH TESTING (ANY CODE REVIEWERS, PLEASE REJECT MERGES TO MASTER IF THIS IS STILL HERE)
+        // TODO: REMOVE THIS WHEN PID CONSTANTS ARE DONE
```
CoolSpy3, line 64:
```suggestion
        // TODO: REMOVE THIS WHEN PID CONSTANTS ARE DONE (ANY CODE REVIEWERS, PLEASE REJECT MERGES TO MASTER IF THIS IS STILL HERE)
```
You don't get to get out of my comments that easily xD Please restore the DO NOT MERGE TO MASTER warning because without it (and quite possibly even with it) this will end up in master.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -15,148 +15,98 @@
...
-    public void setArmVoltage(double volts) {
+        getArmControlOnSmartDashboard();
+    }
+ 
+    public void driveArm(double vel, double accel, boolean usePID) {
+        double armFeedVolts = getKg() * getCoM().getAngle().getCos() + armFeed.calculate(vel, accel);
+        double armPIDVolts = armPID.calculate(armEncoder.getPosition(), goalPosRad[ARM]);
+        if (getArmPos() > ARM_UPPER_LIMIT_RAD || getArmPos() < ARM_LOWER_LIMIT_RAD) {
+            armFeedVolts = getKg() * getCoM().getAngle().getCos() + armFeed.calculate(0, 0);
+        }
```
CoolSpy3: This means if we ever go out of range that the arm will be held in position. Can you add a velocity check so that it moves if we try to move it back into range.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -15,148 +15,98 @@
...
+        // TODO: REMOVE WHEN DONE WITH TESTING (ANY CODE REVIEWERS, PLEASE REJECT MERGES
+        // TO MASTER IF THIS IS STILL HERE)
+        SmartDashboard.putNumber("ArmFeedVolts", armFeedVolts);
+        SmartDashboard.putNumber("ArmPIDVolts", armPIDVolts);
+        double volts = usePID ? armFeedVolts + armPIDVolts : armFeedVolts;
```
CoolSpy3:
```suggestion
        double volts = armFeedVolts +  usePID ? armPIDVolts : 0;
```
Clarity

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -15,148 +15,98 @@
...
+            wristFeedVolts = wristFeed.calculate(getWristPosRelativeToGround(), 0, 0);
+        }
+        // TODO: REMOVE WHEN DONE WITH TESTING (ANY CODE REVIEWERS, PLEASE REJECT MERGES
+        // TO MASTER IF THIS IS STILL HERE)
+        SmartDashboard.putNumber("WristFeedVolts", wristFeedVolts);
+        SmartDashboard.putNumber("WristPIDVolts", wristPIDVolts);
+        double volts = usePID ? wristFeedVolts + wristPIDVolts : wristFeedVolts;
+        SmartDashboard.putNumber("WristTotalVolts", volts);
         wristMotor.setVoltage(volts);
     }
```
CoolSpy3: See my comments on `driveArm`

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -169,11 +119,13 @@ public double getKg() {
...
-        return MathUtil.clamp(MathUtil.inputModulus(armEncoder.getPosition(), ARM_ANGLE_MIN, ARM_ANGLE_MAX), ARM_ANGLE_MIN + ARM_ANGLE_TOLERANCE, ARM_ANGLE_MAX - ARM_ANGLE_TOLERANCE);
+        return MathUtil.inputModulus(armEncoder.getPosition(), ARM_DISCONTINUITY_RAD,
+                ARM_DISCONTINUITY_RAD + 2 * Math.PI);
     }
 
     public double getWristPos() {
-        return MathUtil.clamp(MathUtil.inputModulus(wristEncoder.getPosition(), WRIST_ANGLE_MIN, WRIST_ANGLE_MAX), WRIST_ANGLE_MIN + WRIST_ANGLE_TOLERANCE, WRIST_ANGLE_MAX - WRIST_ANGLE_TOLERANCE);
+        return MathUtil.inputModulus(wristEncoder.getPosition(), ARM_DISCONTINUITY_RAD,
+                ARM_DISCONTINUITY_RAD + 2 * Math.PI);
     }
```
CoolSpy3: Spot the bug

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -202,46 +159,157 @@ public boolean wristAtSetpoint() {
...
+
     public double getWristGoal(GoalPos[] pos) {
         if (isFront) {
-            return pos[isCube].wristPos;
+            return pos[object].wristPos;
         } else {
-            return Rotation2d.fromRadians(pos[isCube].wristPos).unaryMinus().getRadians();
+            return Rotation2d.fromRadians(pos[object].wristPos).unaryMinus().getRadians();
         }
     }
```
CoolSpy3, line 222: Just call `setArmTarget` and `setWristTarget` with the required `GoalPos` object. If anything, you should have no more than a `getGoalPos()` method.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -202,46 +159,157 @@ public boolean wristAtSetpoint() {
...
     public void toggleCube() {
-        isCube++;
-        isCube %= 2;
+        object++;
+        object %= 2;
     }
+
     public void toggleFront() {
         isFront = !isFront;
     }
```
CoolSpy3: Rather than having all these toggle methods and an integer that stores the current state of the toggle, etc. Just implement Uday's requested OI, should be easier.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -202,46 +159,157 @@ public boolean wristAtSetpoint() {
...
+        builder.addDoubleProperty("WisPos", () -> getWristPos(), null);
+        builder.addDoubleProperty("RobotWisPos", () -> getWristPosRelativeToGround(), null);
+        builder.addDoubleProperty("kP: Arm", () -> kP[ARM], x -> kP[ARM] = x);
+        builder.addDoubleProperty("kP: Wis", () -> kP[WRIST], x -> kP[WRIST] = x);
+        builder.addDoubleProperty("kI: Arm", () -> kI[ARM], x -> kI[ARM] = x);
+        builder.addDoubleProperty("kI: Wis", () -> kI[WRIST], x -> kI[WRIST] = x);
+        builder.addDoubleProperty("kD: Arm", () -> kD[ARM], x -> kD[ARM] = x);
+        builder.addDoubleProperty("kD: Wis", () -> kD[WRIST], x -> kD[WRIST] = x);
+    }
 }
```
CoolSpy3, line 277: Add newline at end of file.

CoolSpy3 (changes requested):
Alright, these are the problems I see rn. I also agree with the changes Dean suggested. One note to limit random commits: For the places where I have suggested changes, rather than clicking the Commit Suggestion button, you can instead create a commit batch which *I think* will allow you to commit all the suggestions as a single commit.

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
brettle: I don't think that's needed. Is there a reason you want to do that?

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
brettle: I don't think that's needed. Is there a reason you want to do that?

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
CoolSpy3: It's mostly just because my brain wants to think of the angle as a value from 0-2pi (or I guess -pi - pi) and I figured the code should reflect that. But the more I think about it, the more I would tend to agree that this isn't really necessary, so I'm just going to resolve this comment and the related one for wrist.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
ProfessorAtomicManiac: I thought the toggles is what Uday wanted?

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
ProfessorAtomicManiac: Those values have not been fully determined yet. I was planning at some point to do a test with the rollers to determine the timings of the cube/cone. Currently, they are mutable via SmartDashboard.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
ProfessorAtomicManiac: I ended up removing them entirely and making driveArm the function that actually drives the motors. The only methods that drive the arm/wrist is the driveArm and driveWrist method (Also method names would make more sense).

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
ProfessorAtomicManiac: See my comment about `setArmVoltage` and `setWristVoltage`

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
CoolSpy3: [I don't think so](https://chat.carlmontrobotics.org/team-199/pl/jhxappmjwfbrdcrkehozbzpkke)

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
FriedLongJohns: "This means if we ever go out of range that the arm will be held in position. Can you add a velocity check so that it moves if we try to move it back into range."

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
FriedLongJohns: clamp, not mod

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
```diff
@@ -59,68 +65,82 @@ public static final class Arm {
...
-        MID(new double[]    {-1.5006099006, -2.11308267430}),
-        HIGH(new double[]   {-1.8861790155, -1.86498478435}),
-        HOLD(new double[] {0.f,0.f});
-        */
+    public static final class GoalPos {
 
+        //#region Goal Positions
         // TODO: PUT IN CORRECT POSITIONS
         // These positions are for if the intake/outtake takes place on the front (battery) side of the robot
         // if intake/outtake on back, the "negative" pos will be used
```
brettle: This is not true. First, as of this writing, the positions in CAD correspond to intake on the front side and outtake on the back. Second, if intaking on the back or outtaking on the front, an unrelated set of positions is needed (ie. you can't just negate or reflect the positions from the opposite sided) because the intake roller mech is not symmetric.

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
```diff
@@ -59,68 +65,82 @@ public static final class Arm {
...
         // 0 = CUBE, 1 = CONE
-        public static GoalPos[] LOW = {new GoalPos(40, -0.9225673417), new GoalPos(40, -0.9225673417)};
-        public static GoalPos[] MID = {new GoalPos(40, -1.5006099006), new GoalPos(40, -1.5006099006)};
-        public static GoalPos[] HIGH = {new GoalPos(Units.degreesToRadians(11), -1.8861790155), new GoalPos(40, -1.8861790155)};
-        public static GoalPos[] STORED = {new GoalPos(Units.degreesToRadians(-90), 0), new GoalPos(40, -1.8861790155)};
-        public static GoalPos[] SHELF = {new GoalPos(Units.degreesToRadians(-90), 0), new GoalPos(0, 0)};
-        public static GoalPos[] SUBSTATION = {new GoalPos(Units.degreesToRadians(-90), 0), new GoalPos(0, 0)};
-        public static GoalPos[] INTAKE = {new GoalPos(Units.degreesToRadians(-90), 0), new GoalPos(Units.degreesToRadians(-90), 0)};
+        // {Cone, Cube}
```
brettle: Should say {Cube, Cone}.

src/main/java/org/carlmontrobotics/robotcode2023/commands/SetArmWristPosition.java
FriedLongJohns: is simple this way
only will fix if have time

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
FriedLongJohns: [Outdated]

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
FriedLongJohns: [Fixed]

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
FriedLongJohns: Done.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
FriedLongJohns: [Outdated]

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
FriedLongJohns: Fixed.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
FriedLongJohns: Done.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
FriedLongJohns: Fixed.

src/main/java/org/carlmontrobotics/robotcode2023/commands/SetArmWristPosition.java
brettle: This is actually going to need a fair amount of work but let's leave that until after the merge.

brettle (changes requested):

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
CoolSpy3: Nope, mod is correct

FriedLongJohns: [f6ce832](https://github.com/DeepBlueRobotics/RobotCode2023/pull/6/commits/f6ce8328508e10a846847eb6a8246134207070b9) fix

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -15,148 +15,142 @@
...
+        SmartDashboard.putNumber("ArmTotalVolts", volts);
         armMotor.setVoltage(volts);
     }
-    public void setWristVoltage(double volts) {
+
+    public void driveWrist(double vel, double accel) {
+        double wristFeedVolts = wristFeed.calculate(getWristPosRelativeToGround(), vel, accel);
+        double wristPIDVolts = wristPID.calculate(getWristPos(), goalPosRad[WRIST]);
+        if ((getWristPos() > WRIST_UPPER_LIMIT_RAD && wristFeedVolts > 0) || 
+            (getWristPos() < WRIST_LOWER_LIMIT_RAD && wristFeedVolts < 0)) {
```
CoolSpy3: This logic isn't correct. even if we're trying to move the arm the correct direction, wristFeedVolts may be > 0 due to kG

CoolSpy3 (changes requested):

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
FriedLongJohns: Fixed

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
FriedLongJohns: replaced with vel <> 0

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
FriedLongJohns: These will dissapear with the OI rewrite
need someone to finish it

PR #8, opened 2023-03-18
Teleop Drive

CoolSpy3:


src/main/java/org/carlmontrobotics/robotcode2023/commands/TeleopDrive.java
```diff
@@ -52,34 +55,45 @@ public void execute() {
...
-    deltaT = Robot.robot.getPeriod();
     // Sets all values less than or equal to a very small value (determined by the idle joystick state) to zero.
     // Used to make sure that the robot does not try to change its angle unless it is moving,
-    if (Math.abs(fwd.getAsDouble()) <= Constants.OI.JOY_THRESH) rawForward = 0.0;
-    else rawForward = maxForward * fwd.getAsDouble();
-    if (Math.abs(str.getAsDouble()) <= Constants.OI.JOY_THRESH) rawStrafe = 0.0;
-    else rawStrafe = maxStrafe * str.getAsDouble();
-
-    rotateClockwise = rcw.getAsDouble();
+    double forward = fwd.getAsDouble();
```
roelle, line 60: What are the units of `forward`? Are they changing?

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
```diff
@@ -39,12 +39,12 @@ public static final class Drivetrain {
         public static final double driveGearing = 6.75;
 
         public static final double driveModifier = 1;
-        public static final double wheelDiameterMeters = Units.inchesToMeters(4.0) * 7.36/7.65 /* empirical correction */;
+        public static final double wheelDiameterMeters = Units.inchesToMeters(4.0);
```
brettle: The empirical correction you removed might be important for accurate auto odometry.

src/main/java/org/carlmontrobotics/robotcode2023/commands/TeleopDrive.java
```diff
@@ -52,34 +55,45 @@ public void execute() {
...
-    double rotationMultiplier = slow.getAsBoolean() ? kSlowDriveRotation : 0.55;
-    return new double[] {currentForward * driveMultiplier, currentStrafe * driveMultiplier, -rotateClockwise * rotationMultiplier};
+
+    // ATM, there is no rotational acceleration limit
+
+    // If the above math works, no velocity should be greater than the max velocity, so we don't need to limit it.
+
+    // Quick-Stop
+    if (Math.abs(forward) <= minVelocityMps) currentForwardVel = 0;
+    if (Math.abs(strafe) <= minVelocityMps) currentStrafeVel = 0;
```
brettle: I recommend against doing this because it could cause tipping when decelerating.

brettle (changes requested):

src/main/java/org/carlmontrobotics/robotcode2023/commands/TeleopDrive.java
CoolSpy3: Units of forward are initially a percentage, but are converted to mps following lines 63-64

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
CoolSpy3: Fair, I'll undo the change.

src/main/java/org/carlmontrobotics/robotcode2023/commands/TeleopDrive.java
CoolSpy3: This was added as a driver request. I figured I would port it over to this version. I don't think they thought it was that important tho, so I'll remove it.

src/main/java/org/carlmontrobotics/robotcode2023/commands/TeleopDrive.java
roelle: Not an essential change, but `forward` could be named to reflect what it is e.g. requested next forward velocity(?).

Changing the units/use of the variable is opening an opportunity for bugs.

I don't think there is a bug here... just a note of advice.

PR #18, opened 2023-03-21
Drive to point and distance correction

ProfessorAtomicManiac:
I'm still not done with driving to the nearest goal position, but everything else should be finished. This should contain the distance sensor correction, the drive to point tests, and the drive to point command. The nearest goal position should be relatively simple to implement since I just need to determine positions for the robot to score and have the robot use the driveToPoint command to drive to the nearest goal.


ProfessorAtomicManiac:
Alright I finished drive to nearest goal position.

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
```diff
@@ -32,8 +34,8 @@ public static final class Drivetrain {
 
         //#region Subsystem Constants
 
-        public static final double wheelBase = Units.inchesToMeters(19.75);
-        public static final double trackWidth = Units.inchesToMeters(28.75);
+        public static final double wheelBase = Units.inchesToMeters(20);
+        public static final double trackWidth = Units.inchesToMeters(21);
```
brettle: Why were these changed?

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
```diff
@@ -60,7 +62,7 @@ public static final class Drivetrain {
         // Determine correct turnZero constants (FL, FR, BL, BR)
         public static final double[] turnZero = RobotBase.isSimulation() ?
             new double[] {0, 0, 0, 0} :
-            new double[] {85.7812, 85.0782 , -96.9433, -162.9492};
+            new double[] {135.616, -7.822, 25.137, -39.199};
```
brettle: Why were these changed?

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
```diff
@@ -80,8 +82,8 @@ public static final class Drivetrain {
         public static final double[] drivekP = {1.82, 1.815, 2.015, 1.915};
         public static final double[] drivekI = {0, 0, 0, 0};
         public static final double[] drivekD = {0, 0, 0, 0};
-        public static final boolean[] driveInversion = {false, false, false, false};
-        public static final boolean[] turnInversion = {true, true, true, true};
+        public static final boolean[] driveInversion = {true, true, true, true};
+        public static final boolean[] turnInversion = {false, false, false, false};
```
brettle: Why were these changed.

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
```diff
@@ -105,6 +107,9 @@ public static final class Drivetrain {
         public static final double[] yPIDController = {4, 0.0, 0.0};
         public static final double[] thetaPIDController = {0.10, 0.0, 0.001};
 
+        //public static final double[] tolerance = {Units.inchesToMeters(1), Units.inchesToMeters(1), 5};
+
+
```
brettle: Why was this added?

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
```diff
@@ -113,15 +118,15 @@ public static final class Drivetrain {
...
+        public static final int driveBackRightPort = 14;
 
-        public static final int turnFrontLeftPort = 7;
-        public static final int turnFrontRightPort = 14;
-        public static final int turnBackLeftPort = 6;
-        public static final int turnBackRightPort = 12;
+        public static final int turnFrontLeftPort = 2;
+        public static final int turnFrontRightPort = 15;
+        public static final int turnBackLeftPort = 4;
+        public static final int turnBackRightPort = 13;
```
brettle: Why were these changed.

brettle (commented):
More review coming but clearly this stuff needs to be fixed.

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
```diff
@@ -379,6 +384,39 @@ public RollerMode(double speed, double time, boolean intake) {
...
+            new Pose2d(14.70, 0.5, new Rotation2d(0))
+        };
+        public static Pose2d[] cubeScoringPos = {
+            new Pose2d(1.85, 1.05, new Rotation2d(0)),
+            new Pose2d(1.85, 2.75, new Rotation2d(0)),
+            new Pose2d(1.85, 4.4, new Rotation2d(0)),
+            new Pose2d(14.70, 1.05, new Rotation2d(0)),
+            new Pose2d(14.70, 2.75, new Rotation2d(0)),
+            new Pose2d(14.70, 4.4, new Rotation2d(0)),
+        };
```
brettle: Document units and location of origin. Does it matter whether we are blue or red?

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
```diff
@@ -394,6 +432,10 @@ public static final class Driver {
             public static final int chargeStationAlignButton = Button.kBack.value;
             public static final int resetFieldOrientationButton = Button.kRightBumper.value;
             public static final int toggleFieldOrientedButton = Button.kStart.value;
+            public static final Axis driveToPointButton = Axis.kRightTrigger;
+            // will comment out above once testing done
+            public static final Axis toggleCubeButton = Axis.kRightTrigger;
```
brettle: Should these be the same (even for testing)?

src/main/java/org/carlmontrobotics/robotcode2023/RobotContainer.java
```diff
@@ -48,7 +52,7 @@ public class RobotContainer {
   public final Limelight lime = new Limelight();
   public final Drivetrain drivetrain = new Drivetrain(lime);
   public final Arm arm = new Arm();
-  public final Roller roller = new Roller();
+  public final Roller roller = new Roller(drivetrain);
```
brettle: Recommend having `Roller` just provide the current position of the game piece and moving the logic that requires the drivetrain into the command.

src/main/java/org/carlmontrobotics/robotcode2023/RobotContainer.java
```diff
@@ -82,9 +86,25 @@ public RobotContainer() {
...
+          new DriveToPoint(() -> roller.getNearestGoal(CUBE), drivetrain),
+          new DriveToPoint(roller::correctPosition, drivetrain)
+        ),
+        new SequentialCommandGroup(
+          new DriveToPoint(() -> roller.getNearestGoal(CONE), drivetrain),
+          new DriveToPoint(roller::correctPosition, drivetrain)
+        ),
+        isCube
+      )
+    );
```
brettle: This is complicated enough to warrant a separate command class.

brettle: Also:
1. You should correct the goal and drive to the corrected goal instead of driving to the wrong goal and then driving to correct.
2. In some cases, you need to turn around in order to reach the corrected goal. If you are implementing that, then when you detect that case you need to make sure you are far enough away from the grid to turn around before actually doing so.
3.  The final goal needs to be slightly back from the grid to ensure that the arm can go up without hitting the grid. Not sure whether your current goals take that into account.

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
ProfessorAtomicManiac: this was for the original CorrectToPoint command, but now the CorrectToPoint uses positionTolerance and velocityTolerance

src/main/java/org/carlmontrobotics/robotcode2023/commands/DriveToPoint.java
```diff
@@ -31,16 +32,15 @@ public DriveToPoint(Pose2d targetPose, Drivetrain drivetrain) {
...
+                            new ChassisSpeeds()
                         )
                     ),
                     drivetrain,
                     new HashMap<>()
                 ).getPathCommand(true, true)
             ),
-            new CorrectToPoint(targetPose, drivetrain)
-
+            new CorrectToPoint(targetPose.get(), drivetrain)
```
brettle: This will use the targetPose's value at the time DriveToPoint is constructed, not when it is run. Is that what you want?

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Drivetrain.java
```diff
@@ -138,11 +148,13 @@ public void periodic() {
                 Timer.getFPGATimestamp() - lime.getNTEntry(limelightTransformForPoseEstimation.name().toLowerCase()).getDoubleArray(new double[7])[6] / 1000);
         }
 
-        SmartDashboard.putNumber("Odometry X", getPose().getTranslation().getX());
-        SmartDashboard.putNumber("Odometry Y", getPose().getTranslation().getY());;
+         SmartDashboard.putNumber("Odometry X",
+         odometry.getEstimatedPosition().getTranslation().getX());
+         SmartDashboard.putNumber("Odometry Y",
+         odometry.getEstimatedPosition().getTranslation().getY());
```
brettle: These changes are not needed. getPose() calls odometry.getEstimatedPosition()

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Drivetrain.java
```diff
@@ -39,12 +43,15 @@ public class Drivetrain extends SubsystemBase implements SwerveDriveInterface {
     private boolean fieldOriented = true;
     private double fieldOffset = 0;
 
+    private double testDistX = 0;
+    private double testDistY = 0;
+    private double testRadTheta = 0;
```
brettle: Move to a test command.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Drivetrain.java
```diff
@@ -120,6 +127,9 @@ public Drivetrain(Limelight lime) {
             for(CANSparkMax driveMotor : driveMotors) driveMotor.setSmartCurrentLimit(80);
         }
 
+        SmartDashboard.putNumber("testDistX", testDistX);
+        SmartDashboard.putNumber("testDistY", testDistY); 
+        SmartDashboard.putNumber("testRadTheta", testRadTheta);
```
brettle: Move to a test command's initSendable such that you can actually modify them from smartdashboard.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Drivetrain.java
```diff
@@ -342,6 +354,35 @@ public double[][] getPIDConstants() {
...
+        // Translation2d xTrans = new Translation2d(testDistX, position.getRotation());
+        // Transform2d transformX = new Transform2d(xTrans, new Rotation2d(0));
+        // Pose2d pose = position.plus(transformX);
+        // Translation2d yTrans = new Translation2d(testDistY, position.getRotation());
+        // Transform2d transformY = new Transform2d(yTrans, new Rotation2d(Math.PI/2));
+        // pose = position.plus(transformY);
+        // CommandScheduler.getInstance().schedule(new DriveToPoint(()->pose, this));
+
+    }
+
```
brettle: Move to test command.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Roller.java
```diff
@@ -39,14 +46,11 @@ public boolean runsWhenDisabled() {
...
+    public Roller(Drivetrain dt) {
         led.setLength(ledBuffer.getLength());
         setLedColor(defaultColor);
-        //SmartDashboard.putData(this);
-
-        // TODO: Get proper values for the speeds and timings. For future pull requests,
-        // do not merge if this is not deleted or speeds/timings have not been determined
-        
+        // SmartDashboard.putData(this);
+        this.dt = dt;
```
brettle: See other comment. Roller should not access drivetrain directly. The command to drive to the nearest goal should.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Roller.java
```diff
@@ -89,6 +93,41 @@ public double getGamePieceDistanceIn() {
...
+            // need to shift robot left
+            distanceToMove = dist - ROLLER_WIDTH / 2;
+            translation = new Translation2d(distanceToMove,
+                    initialPose.getRotation().plus(new Rotation2d(-Math.PI / 2)));
+        }
+        Transform2d transform = new Transform2d(translation, new Rotation2d(0));
+        return initialPose.plus(transform);
+    }
+
+    public Pose2d getNearestGoal(boolean mode)
```
brettle: Change mode to isCube for clarity.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Roller.java
```diff
@@ -89,6 +93,41 @@ public double getGamePieceDistanceIn() {
...
+
+        } else if (dist > RIGHT_LIMIT) {
+            // need to shift robot left
+            distanceToMove = dist - ROLLER_WIDTH / 2;
+            translation = new Translation2d(distanceToMove,
+                    initialPose.getRotation().plus(new Rotation2d(-Math.PI / 2)));
+        }
+        Transform2d transform = new Transform2d(translation, new Rotation2d(0));
+        return initialPose.plus(transform);
+    }
```
brettle: This method will probably be rewritten/replaced when you address other comments, but a few notes:
1. You don't seem to account for the width of the game piece.
2. You are always rotating relative to the initial pose, which doesn't sound right.

brettle (changes requested):

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
brettle: The remove the line instead of commenting it out.

ProfessorAtomicManiac:
There have been significant changes to the Roller Subsystem. The changes were made since we need to know what game piece the roller has to determine the distance to shift the robot. Depending on bugs roller may not work anymore so it is necessary to test that before merging.
I resolved all above issues, but here are a few that I still have left unresolved and also progress on them in general:

-"In some cases, you need to turn around in order to reach the corrected goal. If you are implementing that, then when you detect that case you need to make sure you are far enough away from the grid to turn around before actually doing so."
(For now, I will assume that the driver has already rotated the robot to face or face the opposite of the grid before the driver runs the command. This should not be too hard to do with the `RotateToFieldRelativeAngle` command. I don't have the time to work on this so if this feature is necessary please assign someone else @CoolSpy3.)

-"The final goal needs to be slightly back from the grid to ensure that the arm can go up without hitting the grid. Not sure whether your current goals take that into account."
(I will wait for design to get me better positions for the robot)
If there are any additional comments, I can probably resolve them as long as they are minor.

PR #24, opened 2023-03-24
Rotate for limelight

ProfessorAtomicManiac:
Added buttons to rotate the robot 35 degrees so that limelight odometry is more accurate (in more technical details: to avoid "tag flipping"). Drivers would need to press those buttons to face the robot at a curved angle at an apriltag for accurate odometry. Necessary for commands using `DriveToPoint`.

### DIFF
--- a/src/main/java/org/carlmontrobotics/robotcode2023/RobotContainer.java
+++ b/src/main/java/org/carlmontrobotics/robotcode2023/RobotContainer.java
@@ -5,6 +5,7 @@
 package org.carlmontrobotics.robotcode2023;
 
 import static org.carlmontrobotics.robotcode2023.Constants.OI.MIN_AXIS_TRIGGER_VALUE;
+import static org.carlmontrobotics.robotcode2023.Constants.Drivetrain.turnFaceLimelightDeg;
 
 import java.util.HashMap;
 import java.util.function.BooleanSupplier;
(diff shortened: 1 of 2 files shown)


src/main/java/org/carlmontrobotics/robotcode2023/RobotContainer.java
```diff
@@ -90,6 +92,9 @@ private void configureButtonBindingsDriver() {
     new JoystickButton(driverController, Driver.rotateToFieldRelativeAngle90Deg).onTrue(new RotateToFieldRelativeAngle(Rotation2d.fromDegrees(-90), drivetrain));
     new JoystickButton(driverController, Driver.rotateToFieldRelativeAngle180Deg).onTrue(new RotateToFieldRelativeAngle(Rotation2d.fromDegrees(180), drivetrain));
     new JoystickButton(driverController, Driver.rotateToFieldRelativeAngle270Deg).onTrue(new RotateToFieldRelativeAngle(Rotation2d.fromDegrees(90), drivetrain));
+    axisTrigger(driverController, Driver.angleLeftButton).onTrue(new RotateToFieldRelativeAngle(Rotation2d.fromDegrees(MathUtil.inputModulus(drivetrain.getHeadingDeg() - turnFaceLimelightDeg, -180, 180)), drivetrain));
+    axisTrigger(driverController, Driver.angleRightButton).onTrue(new RotateToFieldRelativeAngle(Rotation2d.fromDegrees(MathUtil.inputModulus(drivetrain.getHeadingDeg() + turnFaceLimelightDeg, -180, 180)), drivetrain));
+
```
brettle, line 97: 1. These rotate relative to the current heading, but shouldn't they rotate to absolute field relative angles like the the d-pad buttons?
2. Is the `inputModulus` call needed? Seems like that should already be happening either in the subsystem or in the command.

brettle (changes requested):

PR #25, opened 2023-03-24
Optimize Arm-Wrist Movement

CoolSpy3:


src/main/java/org/carlmontrobotics/robotcode2023/commands/SetArmWristPositionV3.java
```diff
@@ -0,0 +1,155 @@
...
+        if(Math.signum(currentWristPos) == Math.signum(targetWristPos)) {
+            return new SequentialCommandGroup(
+                foldWristToSide(currentWristPos, arm),
+                moveArm(targetArmPos, arm),
+                moveWrist(targetWristPos, arm)
+            );
+        } else {
+            return new SequentialCommandGroup(
+                foldWristToSide(currentWristPos, arm),
+                moveArm(ARM_VERTICAL_POS_RAD - MIN_WRIST_FOLD_POS_RAD, arm),
```
brettle, line 89: Doesn't this always move the arm to a specific position which might be on the opposite side of the robot? Wouldn't it be more efficient to move to the position on the same side as the arm is currently on via a call to `moveArmToSidedSafeWristMovePosition()`?

src/main/java/org/carlmontrobotics/robotcode2023/commands/SetArmWristPositionV3.java
```diff
@@ -0,0 +1,155 @@
...
+    public String name;
+
+    public SetArmWristPositionV3(double targetArmPos, double targetWristPos, Arm arm) {
+        super(() -> {
+            double currentArmPos = arm.getArmPos();
+            double currentWristPos = arm.getWristPos();
+
+            if(Arm.isWristOutsideRobot(currentArmPos, currentWristPos)) {
+                if(Arm.isWristOutsideRobot(targetArmPos, targetWristPos)) {
+                    return moveSidePosIntoRobot(currentArmPos, currentWristPos, targetArmPos, targetWristPos, arm);
```
brettle: This case seems inefficient for the case where the current and target wrist position are on the same side of the robot.

brettle (changes requested):

src/main/java/org/carlmontrobotics/robotcode2023/commands/SetArmWristPositionV3.java
CoolSpy3: My bad, I wrote the if condition backwards

src/main/java/org/carlmontrobotics/robotcode2023/commands/SetArmWristPositionV3.java
CoolSpy3: For this case, the arm is already near vertical, so the side it moves to shouldn't matter that much. Uday said that moving to the back was preferable.

brettle (approved):

PR #30, opened 2023-04-04
Roller rewrite

ProfessorAtomicManiac:
This code should keep the rollers running when intaking a cone, so that the cone does not slip out. The previous `HoldRoller` command does not work since the wheels don't actually spin when the cone is slipped out. However, we also need to keep track of what object this will intake since intaking a cube and having it constantly run will result in the cube getting popped out the front (supposedly, although may not be true since I've heard there were fab changes).

This code was branched off of DriveToPointTest, since that branch already contained code for recording which game objects is in the roller. I got rid of the functionality that records what game pieces is in the rollers (since at the moment it can easily break) and got rid of all the drive to point stuff. However, there might be some residue code that is used for drive to point rather than keeping game pieces inside.


src/main/java/org/carlmontrobotics/robotcode2023/RobotContainer.java
```diff
@@ -144,9 +144,6 @@ private void configureButtonBindingsDriver() {
   private void configureButtonBindingsManipulator() {
     BooleanSupplier isCube = () -> new JoystickButton(manipulatorController, Manipulator.toggleCubeButton).getAsBoolean();
     BooleanSupplier isFront = () -> new JoystickButton(manipulatorController, Manipulator.toggleFrontButton).getAsBoolean();
-    BooleanSupplier isIntake = () -> !isCube.getAsBoolean();
```
CoolSpy3, line 147: Restore Meeeeeeeee (If this was actually intentional plz explain why)

src/main/java/org/carlmontrobotics/robotcode2023/RobotContainer.java
```diff
@@ -158,24 +155,24 @@ private void configureButtonBindingsManipulator() {
...
-        isIntake
+        new RunRoller(roller, RollerMode.INTAKE_CUBE), 
+        new RunRoller(roller, RollerMode.OUTTAKE_CUBE), 
+        isCube // This is really confusing what? isCube is also being used elsewhere so I didn't change it
       ));
     axisTrigger(manipulatorController, Manipulator.rollerIntakeConeButton)
       .onTrue(new ConditionalCommand(
-        new RunRoller(roller, RollerMode.INTAKE_CONE, Constants.Roller.conePickupColor),
-        new RunRoller(roller, RollerMode.OUTTAKE_CONE, Constants.Roller.defaultColor),
-        isIntake
```
CoolSpy3: Use `isIntake` not `isCube` (If this was actually intentional plz explain why)

src/main/java/org/carlmontrobotics/robotcode2023/commands/RunRoller.java
```diff
@@ -48,18 +46,20 @@ public boolean isFinished() {
...
         SmartDashboard.putNumber("SetRoller Time Elapsed (s)", time);
 
-        // if (true)
-        //     return false;
-        return roller.hasGamePiece() == mode.isIntake && time > mode.time;
+        return roller.hasGamePiece() == isIntake() && time > mode.time;
+    }
+
+    public boolean isIntake() {
+        return (mode.obj != GameObject.NONE);
```
CoolSpy3, line 67: Add a short comment explaining that this logic is correct.

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
```diff
@@ -370,28 +375,34 @@ public static final class Roller {
...
         public static class RollerMode {
-            public static RollerMode INTAKE_CONE = new RollerMode(-0.5, 1, true);
-            public static RollerMode INTAKE_CUBE = new RollerMode(0.3, .25, true);
-            public static RollerMode OUTTAKE_CONE = new RollerMode(0.5, .5, false);
-            public static RollerMode OUTTAKE_CUBE = new RollerMode(-0.5, .5, false);
-            public static RollerMode STOP = new RollerMode(0, 0, true);
+            public static RollerMode INTAKE_CONE = new RollerMode(-0.5, .5, GameObject.CONE, conePickupColor);
+            public static RollerMode INTAKE_CUBE = new RollerMode(0.3, .25, GameObject.CUBE, cubePickupColor);
+            public static RollerMode OUTTAKE_CONE = new RollerMode(0.5, .5, GameObject.NONE, defaultColor);
+            public static RollerMode OUTTAKE_CUBE = new RollerMode(-0.5, .5, GameObject.NONE, defaultColor);
```
CoolSpy3, line 380: Add a short comment explaining why these are none and not cube/cone

src/main/java/org/carlmontrobotics/robotcode2023/commands/CorrectToPoint.java
CoolSpy3, line 1: There's no benefit to using a supplier here, the supplier's values are pulled once in the constructor and then never again. Also what's the purpose of this?

src/main/java/org/carlmontrobotics/robotcode2023/commands/DriveToPoint.java
CoolSpy3, line 1: There's no benefit to using a supplier here, the supplier's values are pulled once in the constructor and then never again. Also what's the purpose of this?

src/main/java/org/carlmontrobotics/robotcode2023/commands/RunRoller.java
```diff
@@ -12,31 +11,30 @@
...
         this.mode = mode;
-        this.ledColor = ledColor;
     }
 
     @Override
     public void initialize() {
-        roller.setSpeed(mode.speed);
-        roller.setLedColor(ledColor);
+        if(roller.hasGamePiece() && isIntake()) cancel();
+        if(!roller.hasGamePiece() && !isIntake()) cancel();
```
CoolSpy3: Add a special case for stop.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Roller.java
```diff
@@ -115,4 +120,4 @@ public double getPosition() {
         return motor.getEncoder().getPosition();
     }
 
-}
+}
```
CoolSpy3: Add newline at end of file.

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
```diff
@@ -151,6 +151,11 @@ public static final class Drivetrain {
         public static final double[] velocityTolerance = {Units.inchesToMeters(1), Units.inchesToMeters(1), 25}; // Meters, Meters, Degrees/Second
 
         //#endregion
+
+        //#region Scoring Positions
+        public static final boolean CUBE = false;
+        public static final boolean CONE = true;
+        //#endregion
```
CoolSpy3: Unused

CoolSpy3 (changes requested):

src/main/java/org/carlmontrobotics/robotcode2023/commands/CorrectToPoint.java
ProfessorAtomicManiac: This was code for driveToPoint. I can revert it back for now but from what I remember, it was to dynamically put in positions to drive to. But you are right. If you want to make it work you probably would need a proxycommand.

src/main/java/org/carlmontrobotics/robotcode2023/commands/DriveToPoint.java
ProfessorAtomicManiac: This was code for driveToPoint. I can revert it back for now but from what I remember, it was to dynamically put in positions to drive to. But you are right. If you want to make it work you probably would need a proxycommand.

CoolSpy3: Ye, please revert this as well as `CorrectToPoint`.

src/main/java/org/carlmontrobotics/robotcode2023/commands/RunRoller.java
```diff
@@ -12,31 +11,30 @@
...
         this.mode = mode;
-        this.ledColor = ledColor;
     }
 
     @Override
     public void initialize() {
-        roller.setSpeed(mode.speed);
-        roller.setLedColor(ledColor);
+        if(roller.hasGamePiece() && isIntake()) cancel();
+        if(!roller.hasGamePiece() && !isIntake() && mode.speed > 0) cancel(); // is not stopping the rollers then cancel
```
CoolSpy3: Just do `if(mode.speed > 0)` around the whole block in case someone passes intake true speed = 0; (For some reason)

CoolSpy3 (approved):
I'm reviewing kinda fast today, so I might add comments if I see some random thing later.

Just one small thing. Other than that looks good! Please test before merging.

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
```diff
@@ -358,6 +358,8 @@ public static final class Roller {
         public static final double rollerToleranceRad = 1 / 20 * 2 * Math.PI;
         public static final double rollerHoldSpeedPercent = .3;
 
+        public static final int ROLLER_MAX_CURRENT_AMPS = 20;
```
brettle, line 361: This is fine, but just fyi, that is also the current default.

src/main/java/org/carlmontrobotics/robotcode2023/commands/RunRoller.java
```diff
@@ -12,31 +11,32 @@
...
 
     @Override
     public void execute() {}
 
     @Override
     public void end(boolean interrupted) {
-        roller.setSpeed(0);
+        // keep it running if its a cone
+        if (mode.obj != GameObject.CONE)
+            roller.setSpeed(0);
```
brettle: What happens if there is a cone in the intake and the operator (perhaps accidentally) presses the button for cube intake?

brettle: This is not sufficient to keep the roller running. Think about what will happen when the `RollerCommand` ends.

brettle (changes requested):

brettle (approved):

PR #32, opened 2023-04-28
Safe mode

ProfessorAtomicManiac:
Don't merge into master until after everything is tested out


src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
```diff
@@ -140,7 +140,7 @@ public static final class Drivetrain {
 
         public static final double kNormalDriveSpeed = 1; // Percent Multiplier
         public static final double kNormalDriveRotation = 0.5; // Percent Multiplier
-        public static final double kSlowDriveSpeed = 0.4; // Percent Multiplier
+        public static final double kSlowDriveSpeed = 0.1; // Percent Multiplier
```
ProfessorAtomicManiac: Do not change kSlowDriveSpeed. This is slow mode for when the drivers are driving the robot for more precise control. This is NOT slow mode for when non-trained drivers are using the robot. Make another variable for that but use the same values.

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
```diff
@@ -222,6 +222,8 @@ public static final class Arm {
 
         // TODO: Determine actual max vel/accel
         // public static double[] MAX_FF_VEL = {.25, .25}; // rad / s
+        //Arm, Wrist
+        public static double[] ARM_SLOWMODE_MULT = {.1, .5};
         public static double[] MAX_FF_VEL_MANUAL = {1, 3}; // rad / s
         public static double[] MAX_FF_VEL_AUTO = {1.25, 5}; // rad / s
         public static double[] MAX_FF_ACCEL = {5, 5}; // rad / s^2
```
ProfessorAtomicManiac: `ARM_SLOWMODE_MULT` is redundant. `MAX_FF_VEL_MANUAL` automatically sets the maximum velocity via Trapezoidal control. If you want to limit the speed make another variable for slow mode.

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
```diff
@@ -399,6 +402,14 @@ public RollerMode(double speed, double time, GameObject obj, Color ledColor) {
                 this.obj = obj;
```
ProfessorAtomicManiac: Rollers don't need to be slower for kid mode. The speeds should not change at all. If the rollers are slower it may have trouble picking up game objects. Also by convention we shouldn't be changing the speeds with the `setSpeed()` if the objects represented are constants.

src/main/java/org/carlmontrobotics/robotcode2023/RobotContainer.java
```diff
@@ -132,15 +134,15 @@ public RobotContainer() {
...
       () -> inputProcessing(getStickValue(driverController, Axis.kLeftX)),
       () -> inputProcessing(getStickValue(driverController, Axis.kRightX)),
-      () -> driverController.getRawButton(Driver.slowDriveButton)
+      () -> driverController.getRawButton(Driver.slowDriveButton) || isdriverchild
```
ProfessorAtomicManiac: Again this button is for slow mode for the driver, but NOT baby mode.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -149,9 +150,10 @@ public void autoCancelArmCommand() {
     //#region Drive Methods
 
     private void driveArm(TrapezoidProfile.State state) {
+        double speedMult = RobotContainer.isdriverchild ? ARM_SLOWMODE_MULT[0] : 1;
         double kgv = getKg();
         double armFeedVolts = kgv * getCoM().getAngle().getCos() + armFeed.calculate(state.velocity, 0);
-        double armPIDVolts = armPID.calculate(getArmPos(), state.position);
+        double armPIDVolts = armPID.calculate(getArmPos(), state.position) * speedMult;
```
ProfessorAtomicManiac: I see what you were trying to do here, but just use `MAX_FF_VEL` instead. Switch between those two constraints when switching from baby to normal mode.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -169,9 +171,10 @@ private void driveArm(TrapezoidProfile.State state) {
     }
 
     private void driveWrist(TrapezoidProfile.State state) {
+        double speedMult = RobotContainer.isdriverchild ? ARM_SLOWMODE_MULT[1] : 1;
         double kgv = wristFeed.calculate(getWristPosRelativeToGround(), state.velocity, 0);
         double wristFeedVolts = wristFeed.calculate(getWristPosRelativeToGround(), state.velocity, 0);
-        double wristPIDVolts = wristPID.calculate(getWristPos(), state.position);
```
ProfessorAtomicManiac, line 174: See comment above, and also you are using `ARM_SLOWMODE_MULT` instead of `WRIST_SLOWMODE_MULT`

src/main/java/org/carlmontrobotics/robotcode2023/Robot.java
```diff
@@ -26,6 +27,8 @@ public void robotInit() {
     DriverStation.startDataLog(DataLogManager.getLog());
     if(!DriverStation.isFMSAttached()) PathPlannerServer.startServer(5811);
     robotContainer = new RobotContainer();
+
+    SmartDashboard.putBoolean("safeMode", false);
```
ProfessorAtomicManiac, line 32: @CoolSpy3 i can't think of a better implementation to have a boolean that needs to be accessed everywhere while being constantly updated. I've seen other teams using a `SmartNumber` custom class that does all the initialization and updates for you instead of explicitly throwing it into periodic. Should we do something similar for cases like these?

ProfessorAtomicManiac (commented):
Technically this code has already been tested and it works ok, but I still see a lot of problems with it that I would like to see resolved ideally before Save the Music. I can't "Request Changes" on my own branch but I do not approve this PR. @FriedLongJohns (work on arm related issues) or @BrownBear85 (work on drivetrain related issues) please fix this when you have time.

src/main/java/org/carlmontrobotics/robotcode2023/Robot.java
ProfessorAtomicManiac: Oh wait is that what LiveWindow does with the `initSendable` method. Nvm what I said then

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
```diff
@@ -399,6 +404,14 @@ public RollerMode(double speed, double time, GameObject obj, Color ledColor) {
...
                 this.ledColor = ledColor;
             }
+
+            public double getSpeed() {
+                return speed;
+            }
+
+            public void setSpeed(double speed) {
+                this.speed = speed;
+            }
```
ProfessorAtomicManiac: By convention we shouldn't be changing the speeds with the `setSpeed()` if the objects represented are constants. This also means you don't need a getter either, just make the variable public.

src/main/java/org/carlmontrobotics/robotcode2023/commands/TeleopDrive.java
```diff
@@ -28,18 +28,20 @@ public class TeleopDrive extends CommandBase {
...
   private BooleanSupplier slow;
+  private BooleanSupplier baby;
   private double currentForwardVel = 0;
   private double currentStrafeVel = 0;
 
   /**
    * Creates a new TeleopDrive.
    */
-  public TeleopDrive(Drivetrain drivetrain, DoubleSupplier fwd, DoubleSupplier str, DoubleSupplier rcw, BooleanSupplier slow) {
+  public TeleopDrive(Drivetrain drivetrain, DoubleSupplier fwd, DoubleSupplier str, DoubleSupplier rcw, BooleanSupplier slow, BooleanSupplier baby) {
```
ProfessorAtomicManiac: I really dislike how we are passing in two BooleanSuppliers that do very similar things. Can we merge them together?

src/main/java/org/carlmontrobotics/robotcode2023/commands/TeleopDrive.java
```diff
@@ -67,8 +69,8 @@ public double[] getRequestedSpeeds() {
     if (Math.abs(rotateClockwise) <= Constants.OI.JOY_THRESH) rotateClockwise = 0.0;
     else rotateClockwise *= maxRCW;
 
-    double driveMultiplier = slow.getAsBoolean() ? kSlowDriveSpeed : kNormalDriveSpeed;
-    double rotationMultiplier = slow.getAsBoolean() ? kSlowDriveRotation : kNormalDriveRotation;
+    double driveMultiplier = baby.getAsBoolean() ? kBabyDriveSpeed : (slow.getAsBoolean() ? kSlowDriveSpeed : kNormalDriveSpeed);
+    double rotationMultiplier = baby.getAsBoolean() ? kBabyTurnSpeed : (slow.getAsBoolean() ? kSlowDriveRotation : kNormalDriveRotation);
```
ProfessorAtomicManiac: See my comment about the two boolean suppliers

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -81,53 +90,79 @@ public Arm() {
...
+        SmartDashboard.putNumber("Last Update (s)", lastMeasuredTime);
+        
+        // when the value is different
+        double currentArmPos = getArmPos();
+        if (currentArmPos != lastArmPos) {
+            lastMeasuredTime = currTime;
+            lastArmPos = currentArmPos;
+        }
+        isArmEncoderConnected = currTime - lastMeasuredTime < DISCONNECTED_ENCODER_TIMEOUT_SEC;
+        SmartDashboard.putBoolean("ArmEncoderConnected", isArmEncoderConnected);
```
ProfessorAtomicManiac: Let's get rid of all debugging SmartDashboard code so that the robot doesn't waste memory.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -154,11 +189,9 @@ private void driveArm(TrapezoidProfile.State state) {
         double armPIDVolts = armPID.calculate(getArmPos(), state.position);
         if ((getArmPos() > ARM_UPPER_LIMIT_RAD && state.velocity > 0) || 
             (getArmPos() < ARM_LOWER_LIMIT_RAD && state.velocity < 0)) {
-              forbFlag = true;  
+              forbFlag = true;
```
ProfessorAtomicManiac: If i recall correctly, "forbFlag" was used originally to "rumble" the controller but it never worked. I would not merge unworking code into master so let's get rid of it.

src/main/java/org/carlmontrobotics/robotcode2023/Robot.java
ProfessorAtomicManiac: One of my comments below talks about merging normal, slow, and baby mode into one variable instead of having two variables. You should probably change safeMode to operate similarly.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
```diff
@@ -81,53 +90,79 @@ public Arm() {
...
+        SmartDashboard.putNumber("Current Time", currTime);
+        SmartDashboard.putNumber("Last Update (s)", lastMeasuredTime);
+        
+        // when the value is different
+        double currentArmPos = getArmPos();
+        if (currentArmPos != lastArmPos) {
+            lastMeasuredTime = currTime;
+            lastArmPos = currentArmPos;
+        }
+        isArmEncoderConnected = currTime - lastMeasuredTime < DISCONNECTED_ENCODER_TIMEOUT_SEC;
```
ProfessorAtomicManiac: logic for shutting off the arm when disconnected from encoder should also be copied for wrist

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Arm.java
FriedLongJohns: That's something for the master branch, after all testing is done. It's more useful to waste memory if we aren't in a tournament.

FriedLongJohns:
## Tasks 
`(in chronological order, not importance order!)`

- [x] Copy arm encoder shutoff code to wrist
- [x] Reorganize wrist enum `RollerMode`
- [x] Make sure Roller is unaffected by slow/baby mode
- [x] Review drivetrain baby/slowmode code

*^ Above tasks can be done in any order*

2. - [x] Review that all baby/slow/normal speeds are applied correctly
3. - [x] Merge baby/slow/normal mode into an enum
4. - [x] Charles moment 
5. - [ ] Test code AT WORKSESSION
6. - [ ] Remove debug statements, merge into master

Also, a review of how baby/safemode affects things (Correct me if I'm wrong, @ProfessorAtomicManiac ):
```
  Drivetrain
    slows turning max accel + max vel
    slows movement max accel + max vel
  Arm/Wrist
     slows all feedForward motor accel + max vel
     turns off manual joystick control
   Roller
     no effect
```

ProfessorAtomicManiac:
Yep that's correct!
Also there is a "slow mode" which is completely different from "baby mode". "Slow Mode" does:
```
Drivetrain
  slows turning max accel + max vel (not as much as baby mode i think)
  slows movement max accel + max vel (not as much as baby mode)
Arm/Wrist
   no effect
Roller
   no effect
```

FriedLongJohns (changes requested):
This breaks safe mode AND slow mode.

FriedLongJohns (approved):

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
```diff
@@ -374,32 +380,6 @@ public static final class Roller {
...
-             *               distance sensor has detected an object
-             * @param intake Whether the roller is outtaking or intaking
-             */
-            public RollerMode(double speed, double time, GameObject obj, Color ledColor) {
-                this.speed = speed;
-                this.time = time;
-                this.obj = obj;
-                this.ledColor = ledColor;
-            }
-        }
```
ProfessorAtomicManiac, line 402: Originally Ryan wanted all constants put into Constants.java. Personally, I think constants that are strictly only used in subsystems should stay in their subsystem classes and the only appropriate constants that should be put in Constants.java are constants that are used between subsystems or ports. That way, there is less clutter in Constants.java and it is easier to find certain constants. Seems like you agree with me on this, but do we wanna make it convention that all subsystem specific constants (excluding ports) should stay in subsystems.

src/main/java/org/carlmontrobotics/robotcode2023/Robot.java
```diff
@@ -40,6 +43,14 @@ public void simulationInit() {
...
     CommandScheduler.getInstance().run();
     MotorErrors.printSparkMaxErrorMessages();
+
+    //set mode to baby if baby was true
+		//else but don't set mode to norm if slow was true
+		if (SmartDashboard.getBoolean("safeMode", false)) {
+			RobotContainer.driverMode = RobotContainer.DriverMode.BABY;
+		} else if (RobotContainer.driverMode.isSlow()) {
+			RobotContainer.driverMode = RobotContainer.DriverMode.NORM;
+		}
```
ProfessorAtomicManiac: Once you enable safeMode or slow the robot, how is it supposed to return back to normal mode when switched off?

src/main/java/org/carlmontrobotics/robotcode2023/Constants.java
ProfessorAtomicManiac: Other solutions include creating completely new files just to store constants to organize them better. I realize that storing goal positions in Arm.java would be extremely annoying. We should probably discuss and agree on how exactly we want to store constants.

src/main/java/org/carlmontrobotics/robotcode2023/RobotContainer.java
```diff
@@ -249,17 +294,35 @@ private double inputProcessing(double value) {
...
+    }
+
+    public boolean isSlow() {
+      return this == DriverMode.SLOW;
+    }
+
+    public boolean isBaby() {
+      return this == DriverMode.BABY;
+    }
+  }
```
ProfessorAtomicManiac, line 327: I'm really against having the logic for DriverMode snucked away in a class file meant to initialize joysticks and commands. DriverMode logic is also mixed into Robot.java. In the future if someone wants to change up DriverMode, they would have to edit code in RobotContainer.java and Robot.java, both files that I would argue would not be the first place a programmer would think of to look in. Please refactor the driver mode code (maybe make an entire subsystem class just to keep track of DriverMode, as this is used across multiple subsystems. That's just an idea though, maybe there is a less "boilerplaty" solution).

src/main/java/org/carlmontrobotics/robotcode2023/Robot.java
```diff
@@ -40,8 +44,15 @@ public void simulationInit() {
...
-  }
 
+    // set mode to baby if baby was true
+    // else but don't set mode to norm if slow was true
+    if (SmartDashboard.getBoolean("safeMode", false)) {
+      RobotContainer.driverMode = RobotContainer.DriverMode.BABY;
+    } else if (RobotContainer.driverMode.isSlow()) {
+      RobotContainer.driverMode = RobotContainer.DriverMode.NORM;
+    }
+  }
```
ProfessorAtomicManiac: See my comment about DriverMode in RobotContainer.java

src/main/java/org/carlmontrobotics/robotcode2023/commands/TeleopDrive.java
```diff
@@ -67,8 +69,12 @@ public double[] getRequestedSpeeds() {
...
     else rotateClockwise *= maxRCW;
 
-    double driveMultiplier = slow.getAsBoolean() ? kSlowDriveSpeed : kNormalDriveSpeed;
-    double rotationMultiplier = slow.getAsBoolean() ? kSlowDriveRotation : kNormalDriveRotation;
+    double driveMultiplier = ((DriverMode)mode.get()).isBaby() ? kBabyDriveSpeed : (
+														 ((DriverMode)mode.get()).isSlow() ? kSlowDriveSpeed : kNormalDriveSpeed
+		);
+    double rotationMultiplier = ((DriverMode)mode.get()).isBaby() ? kBabyTurnSpeed : (
+														    ((DriverMode)mode.get()).isSlow() ? kSlowDriveRotation : kNormalDriveRotation
+		);
```
ProfessorAtomicManiac: Reading the code before, I realized that kBabyDriveSpeed was never actually used and baby mode was just "slow mode". This will fix that error but kinda interesting that baby mode was never used.

src/main/java/org/carlmontrobotics/robotcode2023/subsystems/Roller.java
```diff
@@ -118,5 +118,31 @@ public void getRollerConstsOnSmartDashboard() {
...
+         *               distance sensor has detected an object
+         * @param intake Whether the roller is outtaking or intaking
+         */
+        public RollerMode(double speed, double time, GameObject obj, Color ledColor) {
+                this.speed = speed;
+                this.time = time;
+                this.obj = obj;
+                this.ledColor = ledColor;
+        }
+    }
```
ProfessorAtomicManiac, line 147: See my comment in Constants.java

ProfessorAtomicManiac (commented):
From the looks of it, it mostly works. The code is still rough in some places so please try to clean those up before Sunday's worksession.

src/main/java/org/carlmontrobotics/robotcode2023/Robot.java
ProfessorAtomicManiac: Also see my comment in RobotContainer.java

src/main/java/org/carlmontrobotics/robotcode2023/RobotContainer.java
ProfessorAtomicManiac: I also know that Ryan implemented a lib199 implmentation of all the safe mode stuff if you want to use that instead. I forgot how it works though.

